PR #135, opened 2015-02-18
Added MRPT_INCLUDE_DIRS and MRPT_LINK_DIRECTORIES for external usage.

andreasrichtsfeld:
Hi Luis,

I added the variable MRPT_INCLUDE_DIRS and MRPT_LINK_DIRECTORIES for useage in external CMakeLists (e.g. to get it into parent-scope, when having more complex projects).
It would be nice if you could consider that.

Thanks,
Andreas


jlblancoc:
Nice! Seems reasonable and I guess you tested it, so will accept right now. :+1:

PR #167, opened 2015-10-20
rawlog-edit can export more things

mlopezantequera:
--list-poses option works like --list-images, exporting observation poses.
--list-images now also works with 3D range scans (the filenames exported
are those of the intensity images)


jlblancoc:
Thanks!! :+1:

Issue #169, opened 2015-10-29
Camera calibration bug regarding number of distortion coefficients


CBaiz:
Hello,
there is a bug in the camera calibration tool. OpenCV returns either 5 or 8 distortion coefficients (dependent on the flag CV_CALIB_RATIONAL_MODEL), regardless of the size of the passed cv::Mat object. However, in line 310 and 323 in the file checkerboard_cam_calib.cpp, only 4 values are regarded:

...
cv::Mat cameraMatrix, distCoeffs(1,4,CV_64F,cv::Scalar::all(0));
...
for (int i=0;i<4;i++)
     out_camera_params.dist[i] = distCoeffs.ptr<double>()[i];

I've actually checked back with OpenCV to confirm this (see http://answers.opencv.org/question/74243/confusion-about-number-of-distortion-parameters/).
In situations where the 5th parameter is important to the distortion, the calibration suffers from a significant loss of quality (see appended image)

![1](https://cloud.githubusercontent.com/assets/5710585/10830434/d8f545f2-7e7f-11e5-8c8f-e56f856cb18d.png)

jlblancoc:
Thank you very much for the feedback and for checking everything before posting!

Now that you have those images at hand, would you mind modifying the code yourself? You can do a fork and pull request in GitHub. Let me know if this is fine for you...

CBaiz:
I'm trying to build the project with CMake. If I'm successful, I can send a pull request.

CBaiz:
I've successfully fixed the bug. I'll send a pull-request shortly, never done one before.
Also, I wasn't able to debug the application as it always crashed in the function "tokenize" in string_utils.cpp when calling "free(dupStr)", claiming dupStr not to be a pointer to heap-allocated space (which is a lie). 
However, I was able to run a release build and test the bugfix and it's looking good. Comparison (old vs. new):
![versionwithoutbug](https://cloud.githubusercontent.com/assets/5710585/10860371/4f4a750a-7f67-11e5-9e0b-2f4b6f106980.png)
![versionwithbug](https://cloud.githubusercontent.com/assets/5710585/10860370/4f49b07a-7f67-11e5-8929-f99e88c04465.png)

jlblancoc:
Brilliant!! Thanks a lot for the contribution :+1: 

If you want to put your real name & affiliation in the changelog, go and edit it. See last bug fix in https://github.com/MRPT/mrpt/blob/master/doc/doxygen-pages/changeLog_doc.h

[...]

PR #170, opened 2015-10-30
Fix camera calibration bug regarding number of distortion parameters

CBaiz:


jlblancoc:
Fix image rectification bug (k3 may be ignored). 
Closes #169

PR #181, opened 2016-01-21
Fix compilation of CInterfaceFTDI if MRPT_HAD_FTDI is false

astoeckel:
As MRPT_TRY_END is outside the MRPT_HAS_FTDI block, the
MRPT_TRY_START macro should be outside as well.


jlblancoc:
Thanks!

PR #198, opened 2016-03-03
More Python bindings

jlblancoc:


jlblancoc:
@sem23  Hi Peter! I just realized that there were 2 more commits here that were not integrated into MRPT master... 
But before merging, I see something suspicious and wanted to ask you: is it intentional that some classes and files have been DELETED? For example, maps_bindings.h, etc.

jlblancoc:
Will remove this from blocking 1.4.0 until we got a response...

sem23:
Hi JL,

Sorry for beeing that late with my answer. Currently I do not have that
much time to care, because I'm looking for a job, what consumes my whole
time. Hopefully I have some time this weekend to have a look!

Have you ever used them (the bindings), so that you can give me a feedback
if they are useful or not?! :-) And if you think they are useful, please
tell me which classes should be added, to export main functionality to
python! =)

Regards,
Peter

2016-03-09 22:57 GMT+01:00 Jose Luis Blanco-Claraco <
notifications@github.com>:

jlblancoc:
Sorry to hear that, hope you the best luck!!

Yes, I used them only a bit (the pf-localization example,...) and everything seems to work smoothly, so thanks a lot for the contribution! For now it is enough as is, let's the users community request whatever new bindings are needed in the future.

I'll mark this again as blocking 1.4.0 since solving this anytime by ~2 weeks will be more than fine.

NikolausDemmel:
I have used them as well and added some more for my use case. I'll try to provide a pull request and more detailed feedback on the weekend. Thanks already for the addition!

jlblancoc:
Thanks @NikolausDemmel , I look forward your PR!

sem23:
Got some time at weekend. I removed the headers, because there were only
one function prototype inside. I moved the prototypes for each module to
main header. So I think the commit is ok. But now lets go ahead with
debian/ubuntu integration... i prepared something, but it would be nice if
somebody can assist or teach me, how to do it correctly!

2016-03-10 10:44 GMT+01:00 Jose Luis Blanco-Claraco <
notifications@github.com>:

jlblancoc:
Thanks @sem23 !! 

You can take a look at the directories packaging/debian for the current templates. 
It may require a bit a tricking in CMake files as well.. let us know about your progress and doubts on this!

PR #201, opened 2016-03-08
CMake for mac patches

hlzz:
Successfully built mrpt on my mac. However, there seems to be a problem with ffmpeg in other parts of the cmake files. Currently I have to disable mrpt-detectors, mrpt-hwdrivers and mrpt-hmtslam to proceed. Also pymrpt has to be disabled as well since it depends on the above apps.


jlblancoc:
Great, thank you for the contribution!!

PR #202, opened 2016-03-09
Added a new class for clearance grid. Also updated motion_planning_utils

chatrasen:


libs/nav/include/mrpt/nav/tpspace/CParameterizedTrajectoryGenerator.h
```diff
@@ -229,6 +232,42 @@ namespace mrpt
...
+				  : mrpt::utils::CDynamicGrid<mrpt::math::CMatrix>(x_min,x_max,y_min,y_max,resolution),
+				    m_parent(parent)
+				{
+				}
+				virtual ~CClearanceGrid() { }
+
+				bool    saveToFile( mrpt::utils::CStream* fil, const mrpt::math::CPolygon & computed_robotShape );	//!< Save to file, true = OK
+				bool    loadFromFile( mrpt::utils::CStream* fil, const mrpt::math::CPolygon & current_robotShape );	//!< Load from file,  true = OK
+
+				/** For an obstacle (x,y), returns a vector with all the pairs (a,d) such as the robot collides.
```
jlblancoc: @chatrasen This doxygen docs come from a former copy & paste :-)

libs/nav/src/tpspace/CParameterizedTrajectoryGenerator.cpp
```diff
@@ -445,6 +445,17 @@ const CParameterizedTrajectoryGenerator::TCollisionCell & CParameterizedTrajecto
 }
 
 /*---------------------------------------------------------------
+					getTPObstacleClearance
+  ---------------------------------------------------------------*/
+const mrpt::math::CMatrix & CParameterizedTrajectoryGenerator::CClearanceGrid::getTPObstacleClearance(
```
jlblancoc: @chatrasen Returning a "const ref"... it says a lot of your C++ experience, great! ;-)

libs/nav/src/tpspace/CParameterizedTrajectoryGenerator.cpp
```diff
@@ -477,6 +488,30 @@ void CParameterizedTrajectoryGenerator::CColisionGrid::updateCellInfo(
...
+	const uint16_t k,
+	const uint16_t n,
+	const float dist)
+{
+	mrpt::math::CMatrix *matrix = cellByIndex(icx, icy);
+	if(!matrix) return;
+
+	//here we determine whether we need to increase size of matrix or not and then resize the matrix
+	const int cols = std::max(static_cast<int>(matrix->cols()), n+1);
+	matrix->resize(m_parent->getAlfaValuesCount(), cols);
```
jlblancoc: @chatrasen **IMPORTANT** See [this note](http://www.mrpt.org/tutorials/programming/maths-and-geometry/matrices_vectors_arrays_and_linear_algebra_mrpt_and_eigen_classes/#44resizevssetSize) on `resize()`... it will destroy current data!

libs/nav/src/tpspace/motion_planning_utils.cpp
```diff
@@ -136,6 +139,92 @@ void mrpt::nav::build_PTG_collision_grids(
...
+
+				// Robot shape polygon:
+				const mrpt::math::TPolygon2D poly(transf_shape);
+
+				// Get the range of cells that may collide with this shape:
+				const int ix_min = std::max(0,PT->m_clearanceGrid.x2idx(bb_min.x)-1);
+				const int iy_min = std::max(0,PT->m_clearanceGrid.y2idx(bb_min.y)-1);
+				const int ix_max = std::min(PT->m_clearanceGrid.x2idx(bb_max.x)+1,grid_cx_max);
+				const int iy_max = std::min(PT->m_clearanceGrid.y2idx(bb_max.y)+1,grid_cy_max);
+
```
jlblancoc: @chatrasen Here we should have to evalute distances for a larger range... this "+-1" was though for collisions, that is why the check does not extend further. 
How many cells? I think the best would be to offer a new parameter (input argument to this function) defining the maximum clearance in meters to be checked.... it's an idea to start with...

libs/nav/include/mrpt/nav/tpspace/CParameterizedTrajectoryGenerator.h
chatrasen: I didn't get you. These functions would be same for the both the grids right?

libs/nav/src/tpspace/CParameterizedTrajectoryGenerator.cpp
chatrasen: Got your point. So I guess using setsize() would be a safe option here. The padded zeroes won't be a problem as we would access each row only upto npoints.

chatrasen: However, we would be wasting space. Can I use vector of vectors for this purpose? It would be a nice alternative according to me

libs/nav/src/tpspace/motion_planning_utils.cpp
chatrasen: @jlblancoc I will incorporate this change in my next pull request :-)

jlblancoc: :+1:

PR #203, opened 2016-03-09
Incorporated max_clearance and removed other minor bugs

chatrasen:
Instead of resize(), I am using setSize() now. Also, the previous code assumed that both the grids will be written to the same cachefile. I added one more parameter for that.


jlblancoc:
Good! :+1: 
Just before merging: could you please add comments to document that the max_clearance parameter is given in "number of cells" units, to avoid confusion with meters, centimeters, etc.

chatrasen:
@jlblancoc did the commenting part in this commit. Also, have posted a doubt regarding the issue in the comments on the issue itself

PR #206, opened 2016-03-20
Merge remote-tracking branch 'refs/remotes/MRPT/master'

9578577:


jlblancoc:
I think there was some problem here, right? Diff is empty, so I'll close the PR...

PR #251, opened 2016-05-12
Patches for New LibLAS Library

osencan:
**Changed apps/libraries:** 
*  lexical_cast is inserted instead of boost::to_string to convert Project Id to string
* This PR Closes issue on make errors with latest liblas library

I acknowledge to have: 
* Read the [`Discussion`](https://github.com/MRPT/mrpt/commit/aacc1c3a0cbf214e468599fbed5c09c49be7e3b2) page

(Notify: @MRPT/owners )

### DIFF
--- a/libs/maps/src/maps/CPointsMap_liblas.cpp
+++ b/libs/maps/src/maps/CPointsMap_liblas.cpp
@@ -51,7 +51,8 @@ bool CPointsMap::saveLASFile(const std::string &filename, const LAS_WriteParams 
 	const bool has_color = this->hasColorPoints();
 	const float col_fract = 255.0f;
 
-	liblas::Point pt;
+	//liblas::Point pt;
+	liblas::Point pt(&header);
 	liblas::Color col;
 	for (size_t i=0;i<nPts;i++)
 	{
@@ -114,7 +115,8 @@ bool CPointsMap::loadLASFile(const std::string &filename, LAS_HeaderInfo &out_he
 	out_headerInfo.FileSignature      = header.GetFileSignature();
 	out_headerInfo.SystemIdentifier   = header.GetSystemId();
 	out_headerInfo.SoftwareIdentifier = header.GetSoftwareId();
-	out_headerInfo.project_guid       = header.GetProjectId().to_string();
+	//out_headerInfo.project_guid       = header.GetProjectId().to_string();
+	out_headerInfo.project_guid       = boost::lexical_cast<std::string>(header.GetProjectId());
 	out_headerInfo.spatial_reference_proj4 = header.GetSRS().GetProj4();
 	out_headerInfo.creation_year = header.GetCreationYear();
 	out_headerInfo.creation_DOY  = header.GetCreationDOY();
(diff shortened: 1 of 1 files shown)


libs/maps/src/maps/CPointsMap_liblas.cpp
```diff
@@ -113,8 +113,8 @@ bool CPointsMap::loadLASFile(const std::string &filename, LAS_HeaderInfo &out_he
 
 	out_headerInfo.FileSignature      = header.GetFileSignature();
 	out_headerInfo.SystemIdentifier   = header.GetSystemId();
-	out_headerInfo.SoftwareIdentifier = header.GetSoftwareId();
-	out_headerInfo.project_guid       = header.GetProjectId().to_string();
+	out_headerInfo.SoftwareIdentifier = header.
+	out_headerInfo.project_guid       = boost::lexical_cast<std::string>(header.GetProjectId());
```
jlblancoc: @osencan  Just to confirm this: there is a missing `GetSoftwareId();` at the end of your new line 116, right??

jlblancoc:
Closed with commit https://github.com/MRPT/mrpt/commit/804bafd7677a7cbfcf833afbc42d2e20ac1d2021

Thanks!!

libs/maps/src/maps/CPointsMap_liblas.cpp
osencan: Uh! My bad. I did not check after cleaning the commented lines in a remote machine.

jlblancoc: no problem! :+1: 
If possible, please check that everything works after this PR in `master` and let us know if you find further issues...

Cheers!

jlblancoc:
@osencan @jolting Please, take a look (and test, if you are interested) in the changes of commit https://github.com/MRPT/mrpt/commit/c629ef5246c7f5288c81703be009d519420a6888

The comments there are self-explanatory but in short: libLAS imposes such a huge runtime binary+headers dependencies, and we used it only for two functions in MRPT, that it does not make sense to me to enable it by default in Debian/Ubuntu repos.

My proposed solution: as you can see in the commit, move that part to inline functions in a new header which is to be included only by interested users. 

Let me know if you find any issues with this. 
Cheers!

osencan:
@jlblancoc I see that you break the dependencies on libLAS and move all save/load Lidar data features to optional with an external header. Maybe an optional cmake flag like BUILD_BLAS is made this process smoother(with checking dependencies) rather than giving this feature in comments.

Regards,

jlblancoc:
Hi @osencan , 
In later commits I have now re-introduced the CMake-based detection of libLAS. However, I still see more desirable to keep LAS functions in an independent header (as C++ templates) so users can use them independently of whether MRPT was built with or without that library. This is in line with what is done with other 3rd party libs like ZMQ or PCL.

You can see examples of use of the new function in [this line](https://github.com/MRPT/mrpt/blob/master/apps/SceneViewer3D/_DSceneViewerMain.cpp#L1807). Also, the new `#include` is [here](https://github.com/MRPT/mrpt/blob/master/apps/SceneViewer3D/_DSceneViewerMain.cpp#L68). 

Best

PR #261, opened 2016-05-31
Add some trait tests

jolting:
I'll use a pull request to trigger travis. Currently some of the trait tests are failing.

I acknowledge to have: 

(Notify: @MRPT/owners )

Currently some of the PDF pose and point variants are failing.


jlblancoc:
:+1: Great for moving forward 2.0! 

I would prefer to put that into a proper unit test file instead of an example, even if it's build-time only, it will be attempted to build with each `make test`. 
You may have seen them, but just in case, look for `*_unittest.cpp` files under `[MRPT]/libs`. All files following that pattern are moved into unit tests instead of the corresponding lib.

PR #267, opened 2016-06-07
RFC: Add optional intensity to CRangeScan2D

jolting:
I haven't tested this yet, but I wanted to know if you have any suggestions.
Changed: obs
  Add optional intensity to CRangeScan2D

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/libs/obs/include/mrpt/obs/CObservation2DRangeScan.h
+++ b/libs/obs/include/mrpt/obs/CObservation2DRangeScan.h
@@ -70,7 +70,7 @@ namespace obs
 
 		void getScanProperties(T2DScanProperties& p) const;  //!< Fill out a T2DScanProperties structure with the parameters of this scan
 		/** @} */
-		
+
 		/** @name Cached points map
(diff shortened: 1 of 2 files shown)


jlblancoc:
Good idea, seems perfect to me! I can integrate it into master.

I only miss before merging: 
* `hasIntensity` to false at constructor (I think that the standard says that `: hasIntensity()` initializes to `false`, but no mention to `hasIntensity` at all is undefined).
* Extending the `getDescriptionAsText()` method to dump the new vector so it can be, at least, inspected in RawLogViewer for free. 

:+1:  otherwise ;-)

jolting:
@jlblancoc 
hasIntensity is local and should be set by `in >> hasIntensity;`. I key off the !intensity.empty(), so I added a `bool hasIntensity() const`. Also I clear intensity if hasIntensity is false now.

See my latest patch.

jlblancoc:
ok, I see now! Thanks again :+1:

PR #292, opened 2016-07-14
Mrpt var fixes

bergercookie:
**Changed apps/libraries:**
mrpt/base

 COutputLogger_t is a versatile class for logging messages either to the terminal window or to an external file. Class instances can take messages in std::string using the log class methods. Class documentation and usage is summarized in 4f5085d. 
Its main features are also listed below:

- Format of the output message (e.g. in the terminal window) is the following

[LOGGER_NAME | VERBOSITY_LEVEL | CURRENT_TIME] MESSAGE_BODY
- Logger instance stores messages internally so that
   upon request it can dump them either to the console or to an external
   file altogether (this is not the case when the logger.log method is summoned inside const class methods though, see  1b1d05b)
- The message, when printed in the terminal window, is colored according
  to the logger's verbosity/logging level when the message was initialized.
  The available verbosity levels as well as their corresponding colors
  are listed below:

  LVL_DEBUG => CONCOL_BLUE
  LVL_INFO     => CONCOL_NORMAL
  LVL_WARN   => CONCOL_GREEN
  LVL_ERROR => CONCOL_RED

- Filtering of the output messages can be modified (e.g. only show messages of level LVL_WARN or higher) by using the setMinLoggingLevel method.

Output results are also provided:
output_logger
 
I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/libs/base/src/utils/COutputLogger.cpp
+++ b/libs/base/src/utils/COutputLogger.cpp
@@ -136,7 +136,7 @@ void COutputLogger_t::setMinLoggingLevel(const VerbosityLevel& level /*= LVL_INF
 }
 
 
-VerbosityLevel COutputLogger_t::getCurrentLoggingLevel() const { return m_curr_level; }
+VerbosityLevel COutputLogger_t::getLoggingLevel() const { return m_curr_level; }
 
 void COutputLogger_t::getAsString(std::string* fname) const {
 	for (std::vector<TMsg>::const_iterator hist_it = m_history.begin();
@@ -209,7 +209,7 @@ COutputLogger_t::TMsg::TMsg(const std::string& msg_str, const COutputLogger_t& l
 	this->reset();
 
 	name = logger.getName();
-	level = logger.getCurrentLoggingLevel();
+	level = logger.getLoggingLevel();
 	timestamp = mrpt::system::getCurrentTime(); // fill with the current time
 	body = msg_str;
 }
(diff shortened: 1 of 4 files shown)


jlblancoc:
@bergercookie This is just great, thank you very much for the contribution! It was about time to have something like this in MRPT :+1: 

Just a couple of details before accepting the PR:
* Why the `COutputLogger_t` name instead of the shorter and simpler `COutputLogger`? It could be changed...
* Please add a line [like this](https://github.com/MRPT/mrpt/blob/master/libs/base/include/mrpt/utils/CDebugOutputCapable.h#L30) for the doxygen docs grouping. 

A final contribution to make this even more versatile would be adding an alternative signature to `::log()` with a printf format, you can almost directly copy the declaration and implementation from [this](https://github.com/MRPT/mrpt/blob/master/libs/base/include/mrpt/utils/CDebugOutputCapable.h#L40).

Good work!

libs/base/src/utils/COutputLogger.cpp
```diff
@@ -65,49 +74,101 @@ void COutputLogger::log(const std::string& msg_str, const VerbosityLevel& level)
...
-	mrpt::system::sleep(1000);
+	va_list argp;
+	va_start(argp, fmt);
+	std::string str = this->generateStringFromFormat(fmt, argp);
+	va_end(argp);
+
+	this->log(str);
 }
 
+std::string COutputLogger::generateStringFromFormat(const char* fmt, va_list argp) const{
```
bergercookie: Have a common generateStringFromFormat function to be called from both the logFmt **and** logFmt const

jlblancoc: :+1:

jlblancoc:
Ey! I just realized there something missing: please, add the `*_IMPEXP` modifier flag to all classes and structs in MRPT, so they can be correctly exported in Windows (they will work in GCC anyway, but we support Windows too!). 

See an example in the `class ***_IMPEXP Foo` declaration header of *any* class in MRPT, except those which are headers-only.

Issue #295, opened 2016-07-16
FTBFS on sparc64: TopographyReconstructPathFrom3RTK bus error


jlblancoc:
Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=830741

jlblancoc:
List of commits to solve this (for building debian/patch):
* https://github.com/MRPT/mrpt/commit/8dd6537303c6844eeb40e7454edd18e685fbce9b
* https://github.com/MRPT/mrpt/commit/6d6f1001f5a3e31647e751043f49611a7590ffb0
* https://github.com/MRPT/mrpt/commit/6828ef6eec4822a9a0b245cb6ab26a7b5b29f0b3

jlblancoc:
Fixed with PR #298 .
To be submitted to Debian as patched 1:1.4.0-4.

PR #298, opened 2016-07-22
Fix 140 ftbfs sparc64

jlblancoc:
Fix to #295.

PR created from cherry-picking to create a debian/patch file from v1.4.0


jlblancoc:
Closing, as the .patch file has been already generated and uploaded to Debian.

PR #306, opened 2016-08-11
Fix python build

jolting:
@jlblancoc This isn't the most ideal fix because verbosity_level is a reference in TConfigParams. Perhaps verbosity_level could be a shared_pointer when we move to c++11, so that the copy assignment can do the intuitive thing.

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
Thanks Hunter! I thought on some workaround alike, but hadn't time for it since I spent the last 48h travelling... As usual, your help is very appreciated! :+1:

PR #327, opened 2016-09-22
MRPT_INITIALIZER obsoletes do_nothing()

jolting:
I acknowledge to have: 

(Notify: @MRPT/owners )

MRPT_INITIALIZER now prevents the compiler from removing unused global variables


jolting:
@jlblancoc 
I think the constructor for
```
map< std::string , const TSensorClassId *>	CGenericSensor::m_knownClasses;
```
Isn't guaranteed to get called before you attempt to register classes.

I think the easiest way to guarantee this is to treat m_knownClasses like a singleton.

jlblancoc:
Absolutely! :+1: 
I just noticed the error in Travis 5 minutes ago. 
It's already fixed, will test and push...

Issue #312, opened 2016-08-23
graphslam-engine: Integrate with ROS - Add Support for online execution


bergercookie:
@jlblancoc, @EduFdez 
**Goal:** Use the graphslam-engine in an online setup like most of the ROS applciations. Final product should address the following issues:
- Communication of CGraphSlamEngine instance with arbitrary ROS Topics
- Execute graphSLAM (using the already implemented deciders/optimizers in an online fashion - graphslam-engine to provide the deciders/optimisers with the next available action/observations or observation

[Proposed Label] Enhancement
[Proposed Assignee] @bergercookie

bergercookie:
@jlblancoc I've implemented a inter-process communication strategy over TCP Sockets for transmitting measurements fetched from ROS topics (using a ROS node) over to the CGraphSlamEngine application. This enables the graphslam-engine application to be executed in an online fashion in parallel with the measurement acquisition process.

Could you open a new gh topic to which I can PR my changes for further testing and discussion on this?

The branch at hand is [graphslam-devel](https://github.com/bergercookie/mrpt/commits/graphslam-devel)

jlblancoc:
Great news, @bergercookie !! 

I wonder whether you chose this approach instead of wrapping the existing app as a ROS node, avoiding the need to serialize / deserialize in a custom way... but if it's done and it works, it's welcome! ;-)

I created the [new branch](https://github.com/MRPT/mrpt/tree/graphslam-devel) for you to open a PR, but **please** wait until [travis ends](https://travis-ci.org/MRPT/mrpt/builds/163089814) because we're still solving some errors with wx2.8 after another PR earlier today, and I would like yours to start from a clean build...  If travis fails, I'll "rebase" the branch to a future commit when all is ok.

bergercookie:
>I wonder whether you chose this approach instead of wrapping the existing app as a ROS node, avoiding the need to serialize / deserialize in a custom way... but if it's done and it works, it's welcome! ;-)

Indeed, this is a cleaner apprach. However given the current implementation of the CGraphSlamEngine class it is not really feasible as this class contains a parseRawlogFile method that does all the heavy work. 

>I created the new branch for you to open a PR, but please wait until travis ends because we're still solving some errors with wx2.8 after another PR earlier today, and I would like yours to start from a clean build... If travis fails, I'll "rebase" the branch to a future commit when all is ok. 

Ok, I will make the PR after the CI is executed successfully.

Thanks,

[...]

PR #333, opened 2016-09-27
Graphslam devel

bergercookie:
**Changed apps/libraries:** 
- mrpt-graphlam
- graphslam-engine

Current PR provides support for running graphslam-engine application in an online fashion. More specifically, a new abstract class ([CMeasurementProvider](https://github.com/bergercookie/mrpt/blob/7ed106e66c529a16775ba6e4afebed3be7a7b31c/libs/graphslam/include/mrpt/graphslam/CMeasurementProvider.h)) is added that sets the rules for classes that can potentially provide measurements to the CGraphSlamEngine instance. Given this, we have [CRawlogMP](https://github.com/bergercookie/mrpt/blob/7ed106e66c529a16775ba6e4afebed3be7a7b31c/libs/graphslam/include/mrpt/graphslam/CRawlogMP.h) and [CRosTopicMP](https://github.com/bergercookie/mrpt/blob/7ed106e66c529a16775ba6e4afebed3be7a7b31c/libs/graphslam/include/mrpt/graphslam/CRosTopicMP.h) classes that read measurements from rawlog files and ROS topics respectively. 

_note:_ Class CRosTopicMP should be built both on Windows and Linux (despite the fact that ROS practically runs only on the latter) since its not necessary for the MRPT and ROS nodes to coexist on the same machine (e.g. we could have the ROS node - that could run on Ubuntu - listen on a non-localhost address and the MRPT node connects to and receives the required measurements).

_note2:_ Proposed implementation has been successfully tested with LaserScan measurements from a hokuyo laser scanner.

This PR closes issue https://github.com/MRPT/mrpt/issues/312

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,14 +33,14 @@ The following list is roughly sorted in reverse cronological order.
 * Hunter (2016-) https://github.com/jolting
   - Add-ons to 2D range scan observations.
   - Port towards C++11 for MRPT 2.0
-  - Code clean ups. 
+  - Code clean ups.
   - Multiple bug fixes.
-  
+
 * Nikos Koukis, National Technical University of Athens, Greece (2016) - https://github.com/bergercookie
   - Support for executing graphSLAM in graphslam lib/ + Corresponding
   	graphslam-engine application
   - Initial version of MRPT logging module - COutputLogger
-  
+
 * Chandra Mangipudi, University of Minnesota (2016) - https://github.com/mangi020
   - The PnP module in mrpt::vision
 
--- a/apps/graphslam-engine/graphslam-engine_app.cpp
+++ b/apps/graphslam-engine/graphslam-engine_app.cpp
@@ -45,40 +45,40 @@ using namespace std;
 // command line arguments
 // ////////////////////////////////////////////////////////////
 
-TCLAP::CmdLine cmd(/*output message = */ " graphslam-engine - Part of the MRPT\n",
+TCLAP::CmdLine cmd_line(/*output message = */ " graphslam-engine - Part of the MRPT\n",
 		/* delimeter = */ ' ', /* version = */ MRPT_getVersion().c_str());
 
-TCLAP::ValueArg<string> arg_ini_file(/*flag = */ "i", /*name = */ "ini_file",
-		/*desc = */ ".ini configuration file", /* required = */ false,
-		/* default value = */ "", /*typeDesc = */ "config.ini", /*parser = */ cmd);
-TCLAP::ValueArg<string> arg_rawlog_file( "r", "rawlog",
-		"Rawlog dataset file",	false, "", "contents.rawlog", cmd);
+TCLAP::ValueArg<string> arg_ini_file(/*flag = */ "i", /*name = */ "ini-file",
+		/*desc = */ ".ini configuration file", /* required = */ true,
+		/* default value = */ "", /*typeDesc = */ "config.ini", /*parser = */ cmd_line);
 
-// OPTIONAL - If dataset was generated from GridMapNavSimul program and the
-// visualize_ground_truth is set to true in the .ini file, the ground truth is
-// automatically found
-TCLAP::ValueArg<string> arg_ground_truth_file( "g", "ground-truth",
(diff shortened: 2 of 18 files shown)


jlblancoc:
I like much the idea of running it in Windows and still be connected to ros.

The ros side nodes would be a nice contribution to the mrpt-ros-pkgs, in case you are also interested in releasing that part.

Will try it before merging...

jolting:
Is there a special receive node that acts as a bridge between MRPT and ROS? That might lead to a lot of unnecessary complexity.

I was thinking of an alternative that simply uses ros_comm directly, but I haven't quite flushed out how feasible that is. I think ros_comm is portable enough to work across platforms as long as catkin isn't used for building anything. Catkin is great, but I believe Windows has problems with it.
https://github.com/ros/ros_comm/tree/kinetic-devel/clients/roscpp

We don't really need all of ROS. Just supporting the transport is sufficient for basic functionality.

What do you think?

jlblancoc:
> Is there a special receive node that acts as a bridge between MRPT and ROS? That might lead to a lot of unnecessary complexity.

That was what I firstly commented him in #312 : until now, all MRPT functionality that has been ported to ROS follow the same structure, a native ROS node (in a separate repo, not here) making use of the core SLAM/nav/... functionality which is implemented as part of a MRPT lib.

We could do the same here, it should not be difficult: @bergercookie would have to refactor his SLAM app so the most part lies into the mrpt-graphslam library, not in the app (if it's not already like that?).  Then, writing a ROS node using the MRPT lib is not hard. 

But @jolting 's comment made me remember of another alternative: finding and using ROS just like any other dependency. roscomm can be used with FIND_PACKAGE(), without catkin, just like any other lib. I can't remember where, but google'ing you should find some examples. 
Then, MRPT may look for roscomm as an optional component, even in Windows. 

That would be a **much** cleaner and, in the long term, more maintainable solution. The only caveat I can imagine is that we still would need to replicate the same node into the other repo (mrpt-ros-pkg) to be able to make binary releases to the ROS community... but step by step ;-)

Thoughts?

jlblancoc:
I'll accept the PR now, but we can continue using this thread for discussion. 

This is the invaluable repository I remembered where it is very well explained how to use ROS directly from a CMake project (without catkin, package.xml, etc...). I'll try to port that into MRPT cmake scripts to easy @bergercookie 's work in directly listening to ROS nodes from MRPT apps.

jlblancoc:
@bergercookie After these two commits (https://github.com/MRPT/mrpt/commit/bf6bd572ad53a6c35299b10f1cb6a2955c1daae3 and https://github.com/MRPT/mrpt/commit/e2f96a8a1a34acf6294a15f70f396017c0848442 ), MRPT now detects ros_comm and sensors_msgs, etc. as external depedencies. 

It may be difficult to build those two packages in Windows, but if you achieve it, I think it would be a delight! Direct access to a ROS subsystem from an MRPT app in Windows, debugging from MSVC, etc. Let me know what do you think on this possibility.

bergercookie:
> Is there a special receive node that acts as a bridge between MRPT and ROS? That might lead to a lot of unnecessary complexity.

@jolting on the MRPT's side there is a CMeasurementProvider class that in the onlne case sets up a TCP connection. On ROS's side there is a CTopicSniffer class which dumps the measurements on that TCP stream. I don't think it's complicated, especially from the CGraphSlamEngine point-of-view, since all the functionality is included in the CMeasurementProvider class that is picked each time. However I do think that it would be better if it was implemented with standard calls to library methods from the ROS node itself. 

>We could do the same here, it should not be difficult: @bergercookie would have to refactor his SLAM app so the most part lies into the mrpt-graphslam library, not in the app (if it's not already like that?). Then, writing a ROS node using the MRPT lib is not hard. 

Right now I don't know how much time it would take to change the code structure to allow for this, given that I am on a tight schedule with my thesis (multi-robot SLAM) , but I will see to it as soon as possible.

jolting:
Xenial has a ROS client library you can build against. 
http://packages.ubuntu.com/xenial/libroscpp-dev

However, I wouldn't expect anyone to neglect their thesis.

PR #353, opened 2016-10-18
use wxSpinCtrl SetRange method

jolting:
I acknowledge to have: 

(Notify: @MRPT/owners )

wxSpinCtrl does not have a method SetMin or SetMax.
use SetRange instead.
http://docs.wxwidgets.org/trunk/classwx_spin_ctrl.html


jolting:
I haven't tested this, so we should let Jenkins build it first.

jlblancoc:
👍 You're fast, thanks! 

Yes, that method works, it's tested on my local machine. 
For the records: the other methods seem to be new in wx3.1, which is the library version in my dev env...

PR #356, opened 2016-10-20
using rplidar newest SDK 1.5.6 instead of 1.4.3, which support rplidar A1 and rplidar A2

kintzhao:
using rplidar newest SDK 1.5.6 instead of 1.4.3, which support rplidar A1 and rplidar A2
modify the serial port /dev/ttyUSB0  in the configure file of icp-slam-live

[slamtec-rplidar-SDK](http://www.slamtec.com/en/Lidar)
[rplidar_ros](https://github.com/robopeak/rplidar_ros)


jlblancoc:
Ey, thanks a lot! That's a great contribution 👍 

However, Travis says that it failed to build on clang: [see the log](https://travis-ci.org/MRPT/mrpt/jobs/169276985) and search for `error:`.

It would be great if you could address those errors ( [here is how to easily build with clang](http://stackoverflow.com/a/7032021/1631514) ), and even to contribute the corrections back upstream to the rplidar project... 

Let us know if you need help with the process, and thanks again for contributing.

Cheers.

libs/hwdrivers/src/rplidar/src/rplidar_driver.cpp
```diff
@@ -349,10 +570,7 @@ u_result RPlidarDriverSerialImpl::ascendScanData(rplidar_response_measurement_no
...
 
     // all the data is invalid
-    if (i == count){
-        delete[] tmpbuffer;
-        return RESULT_OPERATION_FAIL;
-    }
+    if (i == count) return RESULT_OPERATION_FAIL;
 
     //Tune tail
     for (i = count - 1; i >= 0; i--) {
```
jolting: size_t is unsigned, so this check isn't valid anymore.
Consider
for(rplidar_response_measurement_node_t * ptr = nodebuffer + count - 1; ptr != nodebuffer; ptr--)

libs/hwdrivers/src/rplidar/src/rplidar_driver.cpp
```diff
@@ -298,6 +411,116 @@ u_result RPlidarDriverSerialImpl::_cacheScanData()
...
+        }
+    }
+    _isScanning = false;
+
+    return RESULT_OK;
+}
+
 u_result RPlidarDriverSerialImpl::grabScanData(rplidar_response_measurement_node_t * nodebuffer, size_t & count, _u32 timeout)
 {
     switch (_dataEvt.wait(timeout))
```
jolting: Wait has type of unsigned long. The right thing to do is fix rplidar. wait should return an enum.
This shouldn't be anonymous
```
enum
{
  EVENT_OK = 1,
  EVENT_TIMEOUT = -1,
  EVENT_FAILED = 0,
};
```
This should return an enum
```
unsigned long wait( unsigned long timeout = 0xFFFFFFFF )
```
The quick solution is explicitly cast each one of those case statements into a unsigned long.

libs/hwdrivers/src/rplidar/src/rplidar_driver.cpp
kintzhao: In the whole code block, the logic is ok.  once  all invalid may occure you said, but it is processed at   i== cout.
`    // all the data is invalid
    if (i == count) return RESULT_OPERATION_FAIL;

    //Tune tail
    for (i = count - 1; i >= 0; i--) {
        if(nodebuffer[i].distance_q2 == 0) {
            continue;
        } else {
            while(i != (count - 1)) {
                i++;
                float expect_angle = (nodebuffer[i-1].angle_q6_checkbit >> RPLIDAR_RESP_MEASUREMENT_ANGLE_SHIFT)/64.0f + inc_origin_angle;
                if (expect_angle > 360.0f) expect_angle -= 360.0f;
                _u16 checkbit = nodebuffer[i].angle_q6_checkbit & RPLIDAR_RESP_MEASUREMENT_CHECKBIT;
                nodebuffer[i].angle_q6_checkbit = (((_u16)(expect_angle * 64.0f)) << RPLIDAR_RESP_MEASUREMENT_ANGLE_SHIFT) + checkbit;
            }
            break;
        }
    }`

jolting: Take a look at line 576

kintzhao: it may be a code warning, but not occure error.   I checked it , if "576" invalid,continue i==0, means the data all invalid, equal to i==count,  it does not occur at "576",end at "573".  if have valid data,  for will break at [568](https://github.com/kintzhao/mrpt/blob/master/libs/hwdrivers/src/rplidar/src/rplidar_driver.cpp#L568)

kintzhao: yeah, I agree with you about the warnings. Not notice it without clang and introduce it into sdk at first.  I will try to modify it at next version sdk.

PR #363, opened 2016-10-29
Add load plugin button to ptg-configurator

jolting:
This won't work on Windows and currently there isn't a mechanism for unloading a plugin.

I was thinking about adding it as a command line option,
but I couldn't figure out where the main function is with wxWidgets.

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/apps/ptg-configurator/ptgConfiguratorMain.cpp
+++ b/apps/ptg-configurator/ptgConfiguratorMain.cpp
@@ -12,14 +12,16 @@
 #include "CAboutBox.h"
 
 //(*InternalHeaders(ptgConfiguratorframe)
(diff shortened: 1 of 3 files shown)


jlblancoc:
Simply fantastic! It's more versatile than the other "build-time plug in mechanism", which I can continue using in Windows. 

I'm accepting it. The travis errors are due to the classic annoying string conversions in wxString, I'm fixing it right after merging.

jlblancoc:
For the records: processing cmd-line args in wx can be done like done [here](https://github.com/MRPT/mrpt/blob/master/apps/2d-slam-demo/slamdemoApp.cpp#L91)

PR #367, opened 2016-11-09
Fix undefined reference to symbol on tests linking

corot:
**Changed apps/libraries:**   tests

More precisely: undefined reference to symbol 'pthread_key_delete@@GLIBC_2.2.5

The solution is described [here](http://stackoverflow.com/questions/25617839/undefined-reference-to-symbol-pthread-key-deleteglibc-2-2-5), but seems like nobody bothered to contribute it. Probably someone with more knowledge on CMake can do more elegant

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -43,7 +43,7 @@ ENDIF ()
 
 # Fix issue #367: undefined reference to symbol on tests linking
 IF(NOT MSVC)
-    SET(CMAKE_THREAD_LIBS_INIT ${CMAKE_THREAD_LIBS_INIT} " -lpthread")
+    SET(CMAKE_THREAD_LIBS_INIT ${CMAKE_THREAD_LIBS_INIT} "-lpthread")
 ENDIF()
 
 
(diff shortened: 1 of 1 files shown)


jlblancoc:
Mmm... it's weird that I never had to add that lib manually! 

Anyway, could you please modify it a bit before accepting the PR?
In order to make sure it also works on Windows (MSVC), could you refactor it such that the flag is added after the `endif()` (once instead of twice) and protected inside something like: 
`if(NOT MSVC) ... endif()`
? Thanks!

corot:
Good point... sorry, I sometimes forget that there's live beyond Linux. :see_no_evil: 

Normally, the clean way to go is to add ${CMAKE_THREAD_LIBS_INIT} instead of -pthreads, and CMake fills this variable with whatever is required in your system. But... as explained [here](https://sft.its.cern.ch/jira/browse/ROOT-6585), I found it always empty!!! So I proceeded with the crude -pthreads. That's what I meant when I requested someone good in CMake to do something better.

corot:
Fixed for Windows

jlblancoc:
👍

PR #369, opened 2016-11-11
mrpt-graphslam: Use stream versions to suppress gcc conversion warnings on 32bit systems

bergercookie:
**Changed apps/libraries:** 
mrpt-graphslam

I acknowledge to have: 

(Notify: @MRPT/owners )

Use COutputLogger stream versions instead of logFmt to suppress
warning for conversion from %lu to %u on 32bit systems.

### DIFF
--- a/libs/graphslam/include/mrpt/graphslam/CGraphSlamEngine_impl.h
+++ b/libs/graphslam/include/mrpt/graphslam/CGraphSlamEngine_impl.h
@@ -53,37 +53,35 @@ CGraphSlamEngine<GRAPH_t>::~CGraphSlamEngine() {
 	using namespace mrpt;
 	using namespace std;
 
-	this->logFmt(LVL_INFO, "In Destructor: Deleting CGraphSlamEngine instance...");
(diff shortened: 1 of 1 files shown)


jlblancoc:
I'm having gcc errors due to things like:

	MRPT_LOG_DEBUG_STREAM << ss_out;

where ss_out is a stringstream! 

Please, change them to ss_out.str(), I'm not sure how many times this error happens, but I'm sure you know your code better and can localize the error faster...

jolting:
Would it be easier to just add an operator<< (ostream& )) ?

bergercookie:
@jlblancoc can I reproduce these errors using clang instead of gcc on Ubuntu?

jlblancoc:
I [just added](https://github.com/MRPT/mrpt/commit/e860c0ca3bc4741e8f5793a7c71b6cde1a0c3c3c) a specialization for operator << to handle `stringstream` , so this is solved... 
@jolting if we wanted to add a specialization as you propose, for any ostream object (far more general than stringstream), what would be the equivalent of `s.str()`? Anyway, it's probably okay to leave it as is...

PR #382, opened 2016-11-17
Template <> for pimpl template specialization

jolting:
I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
I know, it should be like that, right?? 
I remember trying it and removing `template <>` in the end because of VS2015 complaining about it... the worst part is that neither gcc nor clang complains, apparently, in travis (!). 
Let's accept your change and see if something happen...

PR #387, opened 2016-11-29
Allow robot poses from external algorithms to be integrated into mrpt particles filter

corot:
**Changed apps/libraries:** 
*   *CLandmarksMap* 
This is a tentative PR following the discussion on https://github.com/mrpt-ros-pkg/mrpt_navigation/issues/43. It's just an addition, so it should not affect current code. The only additions are a likelihood calculation for the robot poses and a new observation class, The idea is to allow the MRPT particle filter to integrate poses estimated by other localization sources, typically not active all the time, for example [ar_track_alvar](http://wiki.ros.org/ar_track_alvar).


jlblancoc:
Great! Please, also add the new class to the registerAll cpp file of the obs library.

I'll take a look in more detail to this and to the original thread asap, before accepting this.

Thanks for contributing!!

jlblancoc:
Merged, thanks!!

PR #400, opened 2016-12-12
Use PF_options.BETA parameter also with adaptive sample size

corot:
**Changed apps/libraries:** 
* SLAM

Now it's only used on fixed sample size, while on adaptive sample size  is hardcoded as 0.5

(Notify: @MRPT/owners )


jlblancoc:
Great... thank you! :+1:

PR #414, opened 2017-01-07
Fix warnings with signed/unsigned int

jolting:
I'm blindly fixing warnings.

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/libs/graphslam/include/mrpt/graphslam/ERD/CICPCriteriaERD_impl.h
+++ b/libs/graphslam/include/mrpt/graphslam/ERD/CICPCriteriaERD_impl.h
@@ -698,7 +698,7 @@ void CICPCriteriaERD<GRAPH_t>::TParams::dumpToTextStream(
 	out.printf("------------------[ Goodness-based ICP Edge Registration ]------------------\n");
 	out.printf("ICP goodness threshold         = %.2f%% \n", ICP_goodness_thresh*100);
 	out.printf("ICP max radius for edge search = %.2f\n", ICP_max_distance);
-	out.printf("Min. node difference for LC    = %d\n", LC_min_nodeid_diff);
+	out.printf("Min. node difference for LC    = %lu\n", LC_min_nodeid_diff);
 	out.printf("Visualize laser scans          = %d\n", visualize_laser_scans);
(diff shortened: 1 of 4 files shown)


jlblancoc:
Cc @bergercookie

bergercookie:
Hey @jlblancoc, @jolting, happy new year.

I thought I had already dealt with most of these warnings. Anyway I will look it up again in the next PR of mine.

Just for reference though, having members as size_t and then printing them with the %lu specifier doesn't work on 32 bit systems.

Cheers.

libs/graphslam/include/mrpt/graphslam/ERD/CICPCriteriaERD_impl.h
```diff
@@ -698,7 +698,7 @@ void CICPCriteriaERD<GRAPH_t>::TParams::dumpToTextStream(
 	out.printf("------------------[ Goodness-based ICP Edge Registration ]------------------\n");
 	out.printf("ICP goodness threshold         = %.2f%% \n", ICP_goodness_thresh*100);
 	out.printf("ICP max radius for edge search = %.2f\n", ICP_max_distance);
-	out.printf("Min. node difference for LC    = %d\n", LC_min_nodeid_diff);
+	out.printf("Min. node difference for LC    = %lu\n", LC_min_nodeid_diff);
```
jlblancoc: Now that you mention it... this is one of those infinity "minor" open fronts in MRPT ;-)

Actually all dumpToTextStream() methods should better use the provided macros, like in:

https://github.com/MRPT/mrpt/blob/29aa46dbb5ae05583407195b6ca5a64dac1594c4/libs/vision/src/CFeatureExtraction_common.cpp#L344

we should then extend the functions in:
https://github.com/MRPT/mrpt/blob/master/libs/base/include/mrpt/utils/CLoadableOptions.h#L33

to support missing types like `size_t`, etc. and after that the correct formatting of all kind of variables should be done only once in `CLoadableOptions.cpp`. A std::stringstream could be used to avoid all the annoying nuisances of printf format specifiers.... they are a real pain ******** to write 100% portable specifiers!

If some of you feel like doing these changes, go on and integrate them with a PR... they have the minimum priority, though...

Cheers!

PR #430, opened 2017-02-04
Minor update to 'm_exprstr_score2_formula' expression

ArashPartow:
This is a very minor change. Initially please note that everything in ExprTk is a value returning entity. As an example the for-loop below will return the value 10 assigning it to x:

`var x := for (var y := 0; y < 10; y +=1 ) { y + 1 }`

Furthermore the _last sub-expression_ is deemed to be the return value of the whole expression.

-------

In the expression **m_exprstr_score2_formula** by adding a _semi-colon ';'_ to the end of the if-statement that creates an end to that particular sub-expression, and implicitly makes the next part the last sub-expression.

I'm thinking the "var result" was added because the expression was resulting in weird values - The reason for that is because there is an implicit multiplication inserted between the if-statement and the final expression. This is what the compiler generates when there isn't the ';' or the `'var result := ...'` :

`if (dif > (num_paths / 2)) { dif := num_paths - dif; } * exp(-abs(dif / (num_paths / 10.0)));`

When the if-statement isn't true and due to there not being an else section (aka alternative) the if-statement returns a **NaN** value.

That NaN is then used in a multiplication operation with the last part of the expression - which then results in "weird" return values - sometimes **NaN**.

Instead by simply adding a _semi-colon_ between the two sub-expressions, the calculation will be as expected and the expression wont incur the run-time cost of evaluating a variable initialization and assignment.

-------

The reason for having implicit multiplication is for scenarios like the following:
- 2x^2
- 2.5abs(3x)
- sin(x)(1/cos(x))
- [Other examples](https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L1066)

### DIFF
--- a/libs/nav/src/reactive/CAbstractPTGBasedReactive.cpp
+++ b/libs/nav/src/reactive/CAbstractPTGBasedReactive.cpp
@@ -83,7 +83,7 @@ CAbstractPTGBasedReactive::CAbstractPTGBasedReactive(CRobot2NavInterface &react_
 	BEST_ETA_MARGIN_TOLERANCE_WRT_BEST(1.05),
 	ENABLE_OBSTACLE_FILTERING(false),
 	m_navlogfiles_dir(sLogDir),
-	m_exprstr_score2_formula("var dif:=abs(k_target-k); if (dif>(num_paths/2)) { dif:=num_paths-dif; } var result := exp(-abs(dif / (num_paths/10.0)));")
+	m_exprstr_score2_formula("var dif:=abs(k_target-k); if (dif>(num_paths/2)) { dif:=num_paths-dif; }; exp(-abs(dif / (num_paths/10.0)));")
 {
 	internal_construct_exprs();
 	this->enableLogFile( enableLogFile );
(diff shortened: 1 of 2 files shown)


jlblancoc:
wow, *thank you* for taking the time to explain it! 

Indeed, I was obtaining a quiet NaN and couldn't understand way, hence the additional `var` "trick"... 

Could you add more examples of usage for `if()` to the library documentation? Most of the time it is used with the if-then-else structure, and I was totally non-confident about what I was doing... Also, the `~{ ...}` structure is used a few times and I didn't know what it means (unless I overlooked it, it seems not to be explicitly explained in the docs! Perhaps examining the source code of the lib, but didn't have much time for it...). 

Thanks again for the great library!

ArashPartow:
> Could you add more examples of usage for if() to the library documentation? Most of the time it is used with the if-then-else structure, and I was totally non-confident about what I was doing...

That's a really good idea. I'll definitely update the documentation, with more examples and an in-depth explanation of the if-statement structure.

jlblancoc:
👍 Thanks!

Issue #434, opened 2017-02-13
CConfigFile improvements


jlblancoc:
To do: 
 - [x] basic preprocessor: split lines with `\` at end of line.
 - [x] Unit tests!

GYengera:
I am new to mrpt and I assumed that this issue would be something I could solve. Could you please explain what exactly needs to be done?

jlblancoc:
Thanks for the offering, @GYengera ! 
I was actually thinking of addressing this in the "short term" (~2 weeks or so, after finishing other topics with higher priority), but this could be indeed a great way for you to get in touch with MRPT so... if you think you can devote time to this in these weeks, it will be fantastic ;-) 

The tasks to do here are: 
 * Create unit tests (read [this](http://www.mrpt.org/libs_tree_layout#6_Unit_testing) and look for examples in existing code) for [CConfigFileMemory](http://reference.mrpt.org/devel/classmrpt_1_1utils_1_1_c_config_file_memory.html). Basic things, as parsing a static `const char *` example configuration file and reading some variables to check if it's parsed right. 
* When done, extend CConfigFileMemory *and* CConfigFile to add a "preprocessor" that allows us to split long config file lines by using the same mechanism than in C/C++ preprocessor: putting a `\` (backslash) *exactly* as the last character in a line. If you are not familiar with this, see [this example](http://stackoverflow.com/questions/10419530/multi-line-preprocessor-macros) (first one I found Googling).

Cheers,

GYengera:
@jlblancoc, I have created a unit test for CConfigFileMemory. However, when I run make test, I only see some of the mrpt unit tests are being run. How do I run the new CConfigFileMemory unit test?

jlblancoc:
I guess you might need to re-run `cmake .` for CMake to detect your new unit tests and include them in the Makefile

GYengera:
The unit test has been run successfully.  I have tested the readString, writeString, getAllKeys and getAllSections functions. How do you want me to send you the code? I will go ahead with the next part now.

jlblancoc:
great! :+1:

You can use the Pull request feature of GitHub.

You should have been working until now in a **fork** of MRPT. If not,
just fork it now and copy the changed files to your local copy
("clone") of that fork. Then commit and push to your fork, and
afterwards open a Pull request against this original repo. Make sure
of selecting the `dev-config-files` branch as target branch, instead
of the default `master`.

PR #450, opened 2017-02-23
CConfigFileMemory unittest

GYengera:
**Changed apps/libraries: Base Utils** 
*  Unit test for CConfigFileMemory

I acknowledge to have: 

(Notify: @MRPT/owners )


libs/base/src/utils/CConfigFileMemory_unittest.cpp
```diff
@@ -0,0 +1,60 @@
...
+   |                                                                           |
+   | Copyright (c) 2005-2017, Individual contributors, see AUTHORS file        |
+   | See: http://www.mrpt.org/Authors - All rights reserved.                   |
+   | Released under BSD License. See details in http://www.mrpt.org/License    |
+   +---------------------------------------------------------------------------+ */
+
+#include <mrpt/utils/CConfigFileMemory.h>
+#include <mrpt/system/os.h>
+
+#include "simpleini/SimpleIni.h"
```
jlblancoc: I think this #include is not required, right? Please, remove it...

libs/base/src/utils/CConfigFileMemory_unittest.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <gtest/gtest.h>
+
+using namespace mrpt;
+using namespace mrpt::utils;
+using namespace mrpt::utils::simpleini;
```
jlblancoc: Idem... not needed from outside of the CConfig* classes... "Users" are not expected to even know about this namespace ;-)

libs/base/src/utils/CConfigFileMemory_unittest.cpp
```diff
@@ -0,0 +1,60 @@
...
+using namespace std;
+
+class TConfig : public CConfigFileMemory {
+	public:
+	using CConfigFileMemory::writeString;
+	using CConfigFileMemory::readString;
+	using CConfigFileMemory::getAllSections;
+	using CConfigFileMemory::getAllKeys;
+	};
```
jlblancoc: I don't understand the motivation of this class here... Why did you need to create it? 
Why not directly define the test objects of type "CConfigFileMemory" instead of TConfig?

libs/base/src/utils/CConfigFileMemory_unittest.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+	vector_string names;
+	TConfig fourth;
+	fourth.writeString("sec","name","val");
+	fourth.writeString("sec","names","value");
+	fourth.getAllKeys("sec", names);
+	EXPECT_STREQ("name",names[0].c_str());
+	EXPECT_STREQ("names",names[1].c_str());
+	EXPECT_EQ(2,names.size());
+}
```
jlblancoc: I would love to see a test, testing the capability of initializing the CConfigFileMemory from a multiline string containing example entries. But I could add it myself after accepting this PR...

libs/base/src/utils/CConfigFileMemory_unittest.cpp
GYengera: The functions within CConfigFileMemory are protected, so the only way I knew of accessing them was through an inherited class. If there is a better way then please do let me know.

jlblancoc: ah, I see!

I just added [this commit](https://github.com/MRPT/mrpt/commit/b6dd211cffd73ec09af01e54321d2e87aadd6f1d) which hopefully will solve your confusion. Those writeString() & readString() are *virtual* methods, not intended for usage from the user's side. 

getAllSections() and getAllKeys() are public, you should have no problem directly using them...

jlblancoc:
👍 

Please, see my further changes in https://github.com/MRPT/mrpt/commit/cd06d5c90ae06bd99f00548884188d6472aa03cf

(You'll have to re-fork or do a PR towards your fork to synch). 

The new feature originally planned in #434 is the one marked as a "TO DO" by the bottom of the file.

PR #456, opened 2017-02-25
Revert resolution change

jolting:
Woops... this is definitely wrong. I was looking at the wrong open.

It seems like the order the streams are opened might have something to do with it, but depth decimation doesn't seem to work automatically in firmware. For now we may be stuck setting all the resolutions the same. Judging by this[1] the firmware is a bit finicky, so it might be worth capturing at a higher resolution and decimating in software to get better range performance.

1. https://github.com/OpenNI/OpenNI2/blob/115cf06c6efea32304182d293eca16ca883c9150/Source/Drivers/PS1080/Sensor/XnSensorIRStream.cpp#L222

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
uf... ok, that's what we have...

PR #459, opened 2017-02-27
Add a extRobotPoseStd parameter to external robot poses likelihood

corot:
**Changed apps/libraries:** 
* SLAM

Add a extRobotPoseStd parameter to external robot poses likelihood calculation, so we can configure their influence in the filter.

**Advice needed:** Actually, I'm not fully sure if this is the right way to calculate the likelihood; I copied the way it's done with beacons, but for example on SIFT is done differently, with an exponential. Suggestions welcomed; thanks!

Also align filter parameters std output.

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
The term is the correct log likelihood of a Gaussian, so it seems okay.

Not sure if it would be "better" (in some way I can't define right now...) to project the error to the SE(3) tangent space (Lie algebra) and evaluate the Mahalanobis distance there... For now let's just go on with this solution, for sure it's good enough!

Thanks... 👍

PR #461, opened 2017-03-01
Completely remove stlplus

jolting:
I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
I **love** this one, thanks!! 👍 

Being thinking a lot on this recently... we should also add definitions for `FooConstPtr` in addition to `FooPtr`, like in PCL, don't you agree?

I also thought (but hasn't tested it yet) that perhaps we could do (via the DECLARE_XXX_POST() macros) 

        using FooConstPtr Foo::ConstPtr;
        using FooPtr Foo::Ptr;

and declare the inner types ::Ptr and ::ConstPtr, with those above for keeping backwards compatibility.

Just some ideas without too much thinking.... it may be impossible to do if we need to forward declare them, though...

jolting:
Personally, I find it difficult to generalize a class when the way it's allocated is so integral to it's definition.

Of course using or typedefs within a class doesn't really increase the complexity much, so I'm for the `FooConstPtr` and `FooPtr` convention.

I'm under the impression that the `DECLARE_XXX_*` macros could be replaced by either a couple of `using` statement with and a templated class. Still thinking about the details though.

Issue #66, opened 2014-07-27
Port to C++14


jlblancoc:
To be done (in 2017? -> yes) when it's safe to assume everyone has a decent C++14 compiler. 

This implies: 
- [x] Deprecate support for Ubuntu precise (and also trusty but only in PPA, for the next item:)
- [x] Use modern CMake 3.1+ C++ compiler features: [target_compile_features](https://cmake.org/cmake/help/v3.1/command/target_compile_features.html#command:target_compile_features), [WriteCompilerDetectionHeader](https://cmake.org/cmake/help/v3.1/module/WriteCompilerDetectionHeader.html), `set (CMAKE_CXX_STANDARD 11)`, etc.
- [x] Smart pointers: forget about stlplus and move to C++11 smart pointers. 
- [x] Threads: move to `std::thread`, unless there are strong reasons not to do so. 
- [x] Switch tfest/indiv-compat-decls.h to `<functional>`
- [x] CMake: use `EXPORT`, etc. 
- [x] Direct use of Eigen? Moved to its own issue: #496 
- [x] Semaphore -> http://stackoverflow.com/a/27852868
- [x] MRPT_OVERRIDE -> `override`
- [x] CPipe : avoid deprecated `auto_ptr<>`
- [x] Search and rewrite all areas having an `#if MRPT_HAS_CXX11`
- [x] Drop `mrpt::synch::CAtomicCounter` in favor of C++11 `std::atomic`.
- [x] Prefer `nullptr`
- [x] `MRPT_NO_THROWS` ==> C++11. 
- [x] Make all destructors `noexcept`
- [x] Consider replacing `uint32_t` enums with correct C++11 typed enums. E.g. in `GNSS_BINARY_MSG_DEFINITION_START`
- [x] Review usages of `mrpt::utils::delete_safe` and consider changing them to use C++11 smart pointers.
- [x] Foo::Create with make_shared ()
- [x] std::function typedefs or similar for ZMQ wrappers: See #231 
- [x] Support move semantics (read [below](https://github.com/MRPT/mrpt/issues/66#issuecomment-219129637))
- [x] `typedef`s ==>  `using`
- [x] Add `constexpr` to geometry constructors, etc.
- [x] `alignas` instead of Eigen align macros
- [x] Simplify ctors via member initialization in headers, e.g. `int a { 0 };` or `int a = 0;`
- [x] Replace `copy_ptr<>`, etc. with simple `using ...` partial template specializations.
[...]

PR #467, opened 2017-03-04
nullptr instead of NULL

olzhas:
*   replaced NULL with nullptr

I am not sure that I did it neatly. I was automatically replacing these things, then after a while I noticed that there are other libraries inside of MRPT, is there a way to remove them?

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
👍 
cc: #66

PR #472, opened 2017-03-12
Add variant type to MRPT based on MapBox

jolting:
Changes to base and added a variant sample.
Included mapbox header only library. Probably will be removed in c++17.

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
wow! I think you must be the unique person on Earth that knows all MRPT corners from A to Z as well or better than myself! :-) 👍 for updating the Debian copyright notice.

Give me some time to test and understand every single line of code here. In general I got the idea and looks fantastic, but want to give it a try with different compilers to be sure it works as expected and does not break anything important... 

Thanks!!

jolting:
>wow! I think you must be the unique person on Earth that knows all MRPT corners from A to Z as well or better than myself! :-) 👍 for updating the Debian copyright notice.

I wouldn't go that far. I just like value types and I wanted to find an alternative to all the `IS_CLASS` and pointer casting stuff.

Plus functional programming is awesome. Maybe this can be a theme for mrpt-2.0 (or 2.1). Map box also has an `optional` class that might be useful, but I haven't looked at it yet.

PR #474, opened 2017-03-13
Add g++-6 and clang-3.8 to travisci

jolting:
Unfortunately, travisci doesn't have a good way to support other versions of compilers.
This doesn't fix the rather verbose warning in Eigen from g++-6. Unfortunately, travis fails the test if the log messages are too big.
Hopefully, the 16.04 version of travis will be available soon and travis will provide better support for switching between C compiler versions.

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,30 @@ sudo: true
 dist: trusty
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then  sudo add-apt-repository ppa:v-launchpad-jochen-sprickerhof-de/pcl -y  ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then  sudo apt-get update -d  ; fi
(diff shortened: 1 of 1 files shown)


jlblancoc:
Ok, thanks!

I this will be useful too for a stalled branch on coverage reports....

jolting:
>I this will be useful too for a stalled branch on coverage reports....

Might not be a bad idea to get that working before GSoC. Students should write unit tests.

jlblancoc:
Sure... 

I think I got coverage working in its branch; at least it worked when executed manually in my system. 

But I'm afraid MRPT is too large to do cov reports via travis: [it just time out](https://travis-ci.org/MRPT/mrpt/jobs/212043317)... this time at 96% complete!! (sigh)

jlblancoc:
FYI: [here](https://coveralls.io/github/MRPT/mrpt?branch=master) is the first valid upload of cov report... done offline and uploaded by manually invoking the coveralls script...

23% coverage, I don't know how accurate is that figure, but in any case it's in the order of magnitude of what I expected...

PR #475, opened 2017-03-13
Removed embedded Eigen3

olzhas:
**Changed apps/libraries:** 
*   Removed embedded Eigen3 from otherlibs/ and added download instructions to CMake building script

I acknowledge to have: 

(Notify: @MRPT/owners )


jolting:
Cool. We'll need to add -Wno-ignored-attributes to CXX_FLAGS on gcc, so that Eigen doesn't complain when using modern g++ versions.

Also the commit I added recently that avoids using the deprecated libav functions needs to be conditional on the libav version, so that we can continue supporting the libav version that ships with Trusty. This can be just an `#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(X, Y, Z)` where X Y Z is the version they deprecated that function.

I'll take a look at that tonight unless you want to give it a shot.

jlblancoc:
Great!! 👍 

Just a note: we will also need to remove the script referenced [here](https://github.com/MRPT/mrpt/blob/master/otherlibs/eigen3/README.docs.txt), which was needed to group Doxygen groups of Eigen under one single namespace, avoiding polluting MRPT doxygen namespaces...

PR #488, opened 2017-03-20
Replace mrpt threads and synchronization primitives

jolting:
MPRT has threads, mutexes, recursive mutexes and semaphores.

Most of these have direct stl replacements.

The only oddball is semaphores, which was replaced by promises.

This removes over 2600 lines from MRPT if you include the test cases and library code.

I acknowledge to have: 

(Notify: @MRPT/owners )


jlblancoc:
I can't overstate **how much** I like this! **Thank you!**

Soon I'll do some tests with programs accessing real hardware after all these changes in the threading functions, but I don't really expect any problem.

jolting:
I imagine a bug or two might show up, but it does knock a couple things off the 2.0 list. Testing is a good idea.

Issue #503, opened 2017-04-02
Integrate `include-what-you-use` in CMake scripts


jlblancoc:
TO-DO: Follow [this idea](http://stackoverflow.com/a/30951493/1631514) to add a new CMake switch that sets the `CXX_INCLUDE_WHAT_YOU_USE` property to all MRPT libs and apps, so this tool can be use with CMake.

PR #507, opened 2017-04-07
Integrate include-what-you-use in CMake script

shivangag:
**Changed CMakeLists.txt:** 
*   Integrated include-what-you-use for CMake version greater than 3.3.
*   Included a switch **USE_IWYU** to turn this option on/off.
*   This PR Closes issue #503 

I acknowledge to have: 

(Notify: @MRPT/owners )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,19 @@ MARK_AS_ADVANCED(FORCE MRPT_ALWAYS_CHECKS_DEBUG)
 OPTION(MRPT_ALWAYS_CHECKS_DEBUG_MATRICES "Additional checks even in Release (Only in matrix classes)" "OFF")
 MARK_AS_ADVANCED(FORCE MRPT_ALWAYS_CHECKS_DEBUG_MATRICES)
 
+# For CMAKE 3.3 and greater, include-what-you-use option
+# ===================================================
+IF(CMAKE_VERSION VERSION_GREATER 3.3)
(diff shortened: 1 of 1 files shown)


jlblancoc:
Thanks! Couldn't tested it yet, but was just curious about the kind of output iwyu gives for mrpt?

jlblancoc:
I have tested this now... I'm afraid there are a couple of mistakes in your patch, e.g. comparing the version against >3.3 instead of >=3.3, and using the GLOBAL_ property, which seems not to be defined, as it is for the target-wise case. I guess you didn't tested the patch, right? 

In any case, I'm closing the PR **and** the issue #507 , since after fixing it, IWYU seems not to work nicely within CMake. It's better to use [this alternative](http://stackoverflow.com/a/31086619/1631514) based on an external script. 

Cheers!

shivangag:
@jlblancoc I actually first tested the script on hwdrivers module using this line
`set_property(TARGET mrpt-hwdrivers PROPERTY CXX_INCLUDE_WHAT_YOU_USE ${iwyu_path})`
and it gave me [this](http://paste.ubuntu.com/24354471/) result.
After getting this result I replace the `TARGET mrpt-hwdrivers` with `GLOBAL` assuming that it includes all libs and applications, which is not the case as I came to know now. Clearly opening a PR without testing it is a big mistake, and I apologize for that.
One thing I noticed while testing is that iwyu gave result when building for the first time only. Therefore to get results I had to delete the old build directory and then build it again.

PR #513, opened 2017-04-23
Minor updates and fixes

ArashPartow:
- Fixes to make ADL easier when using custom defined numeric types
- Added a new production rule as an alias to a previously defined rule


jlblancoc:
Thanks!

ArashPartow:
yw

PR #524, opened 2017-05-26
Allow setting MRPT_DEB_DIR from env variable

jolting:
This is useful if you want builds confined to a workspace (e.g. Jenkins).

I acknowledge to have:

(Notify: @MRPT/owners )


jolting:
I ran into a problem because jenkins-debian-glue uses it's own branch names. Instead of master I changed it to HEAD. Is that right?

jolting:
It might be useful for continuing a maintenance branch in the future, assuming maintenance bug fixes are accepted.

Anyways the snapshot version was taken from HEAD, but the releases were taken from master, so there is a potential mismatch between snapshot version string and the branch you are taking the release from.

PR #527, opened 2017-05-28
Mrpt 2.0 devel

LisGein:
## Changed apps/libraries

* fixed compilation
* added Qt
* added project robot-map-gui

(Notify: @MRPT/owners )


cmakemodules/script_install_commands.cmake
```diff
@@ -45,6 +45,13 @@ IF(WIN32)
 		ENDFOREACH(F)
 	ENDIF (EXISTS "${wxWidgets_LIB_DIR}")
 
+	# Qt5:
+	IF (${CMAKE_MRPT_HAS_Qt5})
+		FOREACH(F ${Qt5_COMPONENTS_LIBS})
```
jolting: Will this cause a problem with Linux?
I think you don't want to do the install of the components on Linux.

jlblancoc:
Good! We'll use this pr to evaluate and test your work during gsoc. I am not sure about whether it should be merged before the end of gsoc, because having a reduced number of PRs helps specifying what was your contribution.

Will review this and test and come back if found some issue.

I totally agree with @jolting, we don't want to install any qt stuff on any platform other than `WIN32`

jolting:
I didn't see the win32 guard. It installs on Linux just fine.
I think this particular PR should be accepted now, so that other projects that depend on Qt can take advantage of this.

PR #535, opened 2017-06-05
Add Quick way of turning off GPL code

jolting:
MRPT is licensed under the BSD-new.
GPL code is compatible with BSD-new.

When sift-hess is built into libmrpt-vision
libmrpt-vision.so is effectively a GPL binary.

ALLOW_GPL=OFF should disable all GPL code.

Carefully seperating the GPL and and BSD-new code
could allow MRPT to release both GPL and BSD-new
licensed binaries as seperate libraries. This
would require a refactoring libmrpt-vision, so that
the sift-hess components are in a seperate library.

I don't imagine that many users are using sift-hess,
so this is good enough for now.

---

I acknowledge to have:

(Notify: @MRPT/owners )


jlblancoc:
Good enough.

PR #538, opened 2017-06-06
Remove PNP so3 due to GPL claim in so3.cpp

jolting:
An SO3 algorithm based on Sophus would be BSD licensed. First step is to remove the GPL one. 

If Chandra P. Mangipudi and Perry Y. Li can remove the GPL3 license text from so3.cpp this might be unnecessary.

---
I acknowledge to have:

(Notify: @MRPT/owners )


jlblancoc:
Thanks. 

I see this a good proposal, so I'll merge it. If @mangi020 , who was the original author of the code, is interested in recovering the PnP SO3 implementation, he can propose a new PR that uses Sophus instead. 

Good 👍

jlblancoc:
BTW: Also for @mangi020 : your implementation of polynomial solvers in [polynom_solver.{h,cpp}](https://github.com/MRPT/mrpt/blob/master/libs/vision/src/pnp/polynom_solver.h) is redundant with that in [mrpt::math](https://github.com/MRPT/mrpt/blob/master/libs/base/include/mrpt/math/poly_roots.h), so it be refactored to use the latter instead...

PR #544, opened 2017-06-14
Modify/correct various elements of build procedure

bergercookie:
See the commit messages for a description of the underlying changes.

---

I acknowledge to have:

(Notify: @MRPT/owners )


CMakeLists.txt
```diff
@@ -485,8 +485,8 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
 	# Use "modern" C99 ! ;-)
  	SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
 
-	# Use modern C++11
- 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+	# Use modern C++14
+ 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
jlblancoc: Ey! I now realize these two lines can be removed, they are redundant with line 21...

cmakemodules/script_pcl.cmake
```diff
@@ -14,10 +14,11 @@ IF(NOT DISABLE_PCL)
 	# PCL library:
 	# --------------------------------------------
 	find_package(PCL COMPONENTS io common registration visualization segmentation surface QUIET)
-	if (PCL_FOUND)
+	if(PCL_FOUND AND PCL_IO_FOUND AND PCL_COMMON_FOUND AND PCL_REGISTRATION_FOUND AND PCL_VISUALIZATION_FOUND AND PCL_SURFACE_FOUND)
+
 		# Filter: It seems clang + c++11 fails to build against PCL (for clang <3.5):
 		if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" AND NOT CMAKE_MRPT_CLANG_VERSION GREATER 34)
```
jlblancoc: Now that I see this... it also fails to build against gcc with c++11 if I recall it right. Try building pbmap targets. 
Well, don't waste your time in it, just wanted to leave it here as a note for future works

jolting: You mean those  SEGFAULTs?

Since GCC 6 c++14 is default, so I'm not too worried about the future. I feel like most of those bugs have gone away in 16.04. I recall having serious issues with 14.04. The issue was binary compatibility with boost[1], not PCL.

Adding PCL to 16.04 requires repackaging of VTK and PCL anyway, so it's unlikely to affect many people. Hopefully VTK can get it right in 18.04, so we can finally have a good PCL package in the Ubuntu repo.

[1] https://lists.boost.org/Archives/boost/2013/05/203762.php

jlblancoc: mmm... nope I referred to linking errors I had in the past while building `mrpt-pbmap`. Errors went like: "symbol not found for  function-foo(std::[c++11]::string...)" (not exact format). 

I always just disabled support for PCL to avoid the problem, which **did seem to me*** to be related to PCL being compiled with non-C++11 std::string so the linker was having problems with different ABIs. 

I never worried too much about it, so this is the first time I googled for it and... found [this](https://gcc.gnu.org/wiki/Cxx11AbiCompatibility) which breaks my theory of why the linker fails (!!). 

Having decent ROS and PCL packages in the official repos will be great... was about time!

PR #552, opened 2017-06-30
Initialize CMetricMapBuilderRBPF properly if map is empty

jolting:
If you try to initialize CMetricMapBuilderRBPF with an empty map
it won't work because it wasn't give it a starting location.

If you try just reinitialize the pose using by calling the other
method.

Also this uses the critical section locker for RTTI.
If an exception is thrown that lock should be unlocked.

---

I acknowledge to have:

(Notify: @MRPT/owners )


jlblancoc:
All good. 
Thanks! Will propagate it to 1.5 since it's a potential bug fix.

PR #556, opened 2017-07-06
ClangFormat - Major reformat of MRPT codebase style

bergercookie:
Hey @jlblancoc, @jolting,

This is the reformatting PR I was referring to in https://github.com/MRPT/mrpt/issues/520. Do take your time to examine the changes to the codebase. Also let me know if you disagree or would like to suggest changes to the applied code style. It's configuration can be found [here](https://github.com/bergercookie/mrpt/blob/mrpt-2.0-devel/scripts/clang_git_format/.clang-format). The majority of the changes are outlined either in https://github.com/MRPT/mrpt/issues/520 or in the commit messages of this PR.

P.S. I 've got a few utility scripts to modify to assist porting of code prior to the reformat so let's keep this PR open for at least the next couple of days.

(Notify: @MRPT/owners )


jlblancoc:
400,000 lines of code change with this... sure, I will take **some time** to give it a look, hahaha! ;-)

Thanks!!

jlblancoc:
Is there any "standard" path for the `.clang-format` file for editors / IDEs to use it? I'm mentioning because it's not at root...

jolting:
clang-format will will recursively search backwards through the directories for .clang-format. Best to put it in the root.

bergercookie:
I just rebased my history and updated the documentation on how to port code to the new ClangFormat style. However the pull-request doesn't seem to pick up the new commits. I am opening a new one..

bergercookie:
>Is there any "standard" path for the .clang-format file for editors / IDEs to use it? I'm mentioning because it's not at root...

I placed a symlink of the scripts/clang_git_format/.clang-format in the MRPT root

jolting:
This has to compile on Windows. I know Windows has *some* support for symlinks, but I wouldn't want to depend on it.

jlblancoc:
On Fri, Jul 7, 2017 at 2:27 AM, Hunter Laux <notifications@github.com> wrote:
> This has to compile on Windows. I know Windows has some support for symlinks, but I wouldn't want to depend on it.

100% agree!

I know throwing more and more files at "root" is undesirable but... if
it's needed, let's put it there and live with it!

.travis.yml
```diff
@@ -48,9 +48,11 @@ install:
     fi
   - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" = "g++" ]; then export CXX="g++-6"; fi
   - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$CXX" == "clang++" ]; then export CXX="clang++-3.8"; fi
+  - pip install -r travis/python_reqs.txt
 
 script:
-  - bash .travis.sh
+  - bash travis/check_style.sh
+  - bash travis/travis_main.sh
```
jlblancoc: You should also change a reference to the old `travis.sh` file in the shippable script (in the root directory)...

jlblancoc: Ah! You already done it, sorry! :-) The search didn't work in the GH UI due to the very large diff... Forget the last comment!

bergercookie: I was sure I dealt with it when I saw the shippable build error. :)

https://github.com/MRPT/mrpt/pull/556/commits/f2834a3b1bdb519d8c2718066314bc0c89dd7d5d

jlblancoc:
What do you think about applying the `do { ... } while (0)` trick to allow/enforce all macros to be ended with a semicolon? It would apply to MRPT_START, MRPT_END, ASSERT_*(),  and so on. 
See examples of use [here](https://github.com/MRPT/mrpt/blob/master/libs/base/include/mrpt/utils/COutputLogger.h#L313). 
I could propose a PR to your fork... or we could add it **after** accepting this PR. 

I mention this now because in some IDEs, the missing ";" confuses the automatic indentation...

Anything related to this idea that implies a change in the clang format file, @bergercookie ??

Thanks!

bergercookie:
>What do you think about applying the do { ... } while (0) trick to allow/enforce all macros to be ended with a semicolon? It would apply to MRPT_START, MRPT_END, ASSERT_*(), and so on.
See examples of use here.

Having read [this](https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros) I think this would be super useful. I don't think there is a ClangFormat option for this though, so all the MRPT macros should be manually edited.

>I could propose a PR to your fork... or we could add it after accepting this PR.

Whatever suits you is fine by me.

>I mention this now because in some IDEs, the missing ";" confuses the automatic indentation...

That's exactly the case with Vim as well! Until now I was manually applying the ending colon... 

>Anything related to this idea that implies a change in the clang format file, @bergercookie ??

Hmm, can't think of any right now..

jlblancoc:
Well, it's physically impossible to review the 7e5 lines of code added/removed by this PR, so, as long as Travis passes, it's OK for me to merge it... 

@bergercookie , @jolting : something to add before doing it? Specially any desire to change the clang format, since each change might have a significant impact in the "weight" of the Git commit tree...

doc/ClangFormat.md
```diff
@@ -80,6 +119,8 @@ TODO
 
 ## TODOs
 
+- [X] Add travis test file
+- [X] Enable to lint one's work offlien - prior to PR/commit
 - [ ] TODO - update the links
 - [ ] Talk about the reformat_branch
 - [ ] Solve all TODOs (how meta! :-) )
```
jlblancoc: What links do you refer here to?

> how meta! 

XDD

jlblancoc:
The only change I feel it's like moving backwards is the change from `//!< XX` to `/** XX */`. Your script seems to have done an amazing work and all should work nicely for Doxygen, but my concern is about **readability** of source code, especially headers with lots of member variables or short methods. 

Using the `//!< XX` notation helped a lot to avoid using **twice** lines of code to specify a class/struct declaration. Given the size of MRPT, I think it's relevant both for the size of the source code, and the ease to visually inspect a header and get an idea of everything there at a glance. 

Is there any particular problem of clang-format with this kind of notation? Anything we could work to fix? 

If yes, you could rebase your branch to remove that commit, I guess, instead of creating another "undo" commit. 

Thoughts?

jlblancoc:
> Having read this I think this would be super useful. I don't think there is a ClangFormat option for this though, so all the MRPT macros should be manually edited.
>> I could propose a PR to your fork... or we could add it after accepting this PR.
> Whatever suits you is fine by me.

I think I'll do that **after** merging this one, so we work on one thing at a time ;-)

doc/ClangFormat.md
bergercookie: >What links do you refer here to?

Links of this sorts:

```sh
[check_style.sh script](TODO - Add link)
```

But its easier to update them after merging this because then I'll know either the exact URL for the file or the commit ID in similar cases...

bergercookie:
>The only change I feel it's like moving backwards is the change from //!< XX to /** XX */. Your script seems to have done an amazing work and all should work nicely for Doxygen, but my concern is about readability of source code, especially headers with lots of member variables or short methods.

>Using the //!< XX notation helped a lot to avoid using twice lines of code to specify a class/struct declaration. Given the size of MRPT, I think it's relevant both for the size of the source code, and the ease to visually inspect a header and get an idea of everything there at a glance.

Hmm.. to tell you the truth I am more in favor of the `/** ... */` notation since the `//!< ...` almost always gets over the 80something lines limit that I set. But if you think we should roll this back then it's fine by me :)

>Is there any particular problem of clang-format with this kind of notation? Anything we could work to fix?

The only thing I see is that, in this case we should completely disable the `ColumnLimit` option which is now set to 80 chars.

>If yes, you could rebase your branch to remove that commit, I guess, instead of creating another "undo" commit.

Sure, I can do that...

jlblancoc:
> Hmm.. to tell you the truth I am more in favor of the /** ... */ notation since the //!< ... almost always gets over the 80something lines limit that I set. But if you think we should roll this back then it's fine by me :)

Ah, that is...

Ok, in that case it makes sense. I'll give it one more check and will
merge soon.

JL

PR #577, opened 2017-08-14
bug(icp-slam): can not run correctly under windows

tyuownu:
The initial pose should be [x = 0, y = 0, phi = 0], but under windows the x, y and phi are all -9.25596e+61, so we need to give them initial value.

## Changed apps/libraries

* modified-libX
* modified-appY
* ...

## PR Description

* (*REMOVE THIS LINE IF NOT NEEDED*) This PR Closes issue #XXX
* (*REMOVE THIS LINE IF NOT NEEDED*)

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/slam/src/slam/CMetricMapBuilderICP.cpp
+++ b/libs/slam/src/slam/CMetricMapBuilderICP.cpp
@@ -178,7 +178,7 @@ void CMetricMapBuilderICP::processObservation(const CObservation::Ptr& obs)
 		// Current robot pose given the timestamp of the observation (this can
 		// include a small extrapolation
 		//  using the latest known robot velocities):
-		TPose2D initialEstimatedRobotPose;
+		TPose2D initialEstimatedRobotPose(0,0,0);
 		{
 			mrpt::math::TTwist2D robotVelLocal, robotVelGlobal;
 			if (obs->timestamp != INVALID_TIMESTAMP)


jlblancoc:
Hi @tyuownu !

I'll answer later to your other question on why TPose2D is not initialized to zeros. 

But regarding your patch, I can't see what's wrong: please, let us know if the wrong initial pose you are having is using the "icp-slam" application, or some other custom application you wrote using MRPT. 

I'm asking this because `initialEstimatedRobotPose` is left uninitialized in that line, but immediately after that, there is an `if` and in both branches (then / else), the value of `initialEstimatedRobotPose` is initialized using other methods. So, if you really observed uninitialized values **after** returning from `processObservation()`, that may be a bug, but somewhere else, because any of those two branches **must** initialize the variable.

tyuownu:
Hi @jlblancoc . Today, I test the `mrpt_icp_slam_2d` program with the rawlog(only scan data).
```
roslaunch mrpt_icp_slam2d icp_slam_myrawlog.launch
```
The car did not move and than I found that the icp do not convergence. Finally, I dig out that it still the default constructor problem. And the `initialEstimatedRobotPose` initial to `[x = 0.000000, y = 0.000000, z = 461486422375072255280957843814205172...]`.
And then, I ran the program with the same configure file and rawlog file:
```
icp-slam icp_slam_demo.ini myrawlog.rawlog
```
The car can move and the `icp algorithm` can run correctly, it's wired. I still do not know the difference, do you have any idea? Maybe you can test it with the attachment below.

I know you purpose about the default constructor about `TPose2D`, but as to this special case, I think it better give the `zeros-value` to `initialEstimatedRobotPose`, what do you think?

```
environment:
OS: ubuntu 16.04 amd64
ROS: Kinetic
MRPT: 1.5.0(I compile it by my self, release type)
```
[attachment files](https://www.dropbox.com/sh/i27u9dqx75ncdqv/AAAjeDN-KlMLocMAy3n4TP7ja?dl=0)

jlblancoc:
I can confirm that, as far as for mrpt master (1.9.9), the app `icp-slam` works correctly with yout INI + rawlog. 

So, if I understood you correctly: 
* when running with the ROS node for icp-slam, it doesn't work? 
* have you verified that, by simply changing the inicialization to (0,0,0), the ROS node is fixed? (??)

(I can't test myself the ROS nodes right now)

tyuownu:
Hi @jlblancoc .
1. When running ROS node `mrpt_icp_slam_2d` with my rawlog file, it doesn't work.
without the commit
The ICP algorithm produce a low quality.

2. And changing the initialization, it works.
screenshot from 2017-08-18 09-30-55

tyuownu:
I also test my INI and rawlog file with `icp-slam`, and it's works. But not work on the `mrpt_icp_slam_2d`, do you have some idea? (seems `mrpt_icp_slam_2d` and `icp-slam` give difference default value to `initialEstimatedRobotPose`).

jlblancoc:
Thank you for your tests!

I will dig into this because there must be another bug somewhere. 

After that, I'll include your patch anyway, and it will go into master and the next 1.5.4 release.

jlblancoc:
@tyuownu Please, could you check if: 
* leaving the TPose2D back again to uninitialized (that is, remove the (0,0,0) inialization you added), and 
* applying this commit https://github.com/MRPT/mrpt/commit/1b87f426fd9ffab531c02505f0be89ab362867b9

solves the issue as well? Thanks.

Even if it works, I'll apply the (0,0,0) initialization since its costs is minimal, but want to be sure what caused this error...

tyuownu:
@jlblancoc . OK, I will test it tomorrow, and tell you the result.

tyuownu:
@jlblancoc . I test your patch, and it's works fine.
BTW, I found that, the 'initialEstimatedRobotPose' initial not always that `phi` equal to a undefined value, sometime, `x` to that undefined value too.
Like:
`[x = 0.000000, y = 0.000000, phi = 461486422375072255280957843814205172...]`, sometime,
`[x = 461486422375072255280957843814205172..., y = 0.000000, phi = 0.000000]`.

tyuownu:
cheers!

PR #580, opened 2017-08-21
General improvements and fixes

ArashPartow:
Added fixes for a clean GCC 7.+ compile

### DIFF
--- a/libs/base/include/mrpt/otherlibs/exprtk.hpp
+++ b/libs/base/include/mrpt/otherlibs/exprtk.hpp
@@ -67,6 +67,20 @@ namespace exprtk
    #define exprtk_error_location             \
    "exprtk.hpp:" + details::to_str(__LINE__) \
 
+   #if __GNUC__  >= 7
+
+      #define exprtk_disable_fallthrough_begin                      \
+      _Pragma ("GCC diagnostic push")                               \
+      _Pragma ("GCC diagnostic ignored \"-Wimplicit-fallthrough\"") \
+
+      #define exprtk_disable_fallthrough_end                        \
+      _Pragma ("GCC diagnostic pop")                                \
+
+   #else
+      #define exprtk_disable_fallthrough_begin (void)0;
+      #define exprtk_disable_fallthrough_end   (void)0;
+   #endif
+
    namespace details
    {
       typedef unsigned char uchar_t;
@@ -1666,6 +1680,7 @@ namespace exprtk
 
          if (length <= 4)
          {
+            exprtk_disable_fallthrough_begin
             switch (length)
             {
                #ifdef exprtk_use_lut
@@ -1699,6 +1714,7 @@ namespace exprtk
 
                #undef exprtk_process_digit
             }
+            exprtk_disable_fallthrough_end
          }
          else
             return_result = false;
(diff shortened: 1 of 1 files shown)


jolting:
I'd prefer using the `[[fallthrough]]` attribute when using gcc 7 instead of `-Wno-`... flags. We might want to conditionally enable c++17 when using gcc 7 assuming the falltrhough attribute requires c++17 on gcc 7.

http://en.cppreference.com/w/cpp/language/attributes

jlblancoc:
Thanks @ArashPartow ! 

@jolting : Since we now enable C++XX via `CMAKE_CXX_STANDARD` at the very beginning of the main CMake script instead of specific compiling flags, I'm not sure if it might cause any issue to delay the declaration of `CMAKE_CXX_STANDARD` until the compiler has been detected. Probably not... 

Anyway, a less aggressive & less global change, would be to enable C++17 just in the unique file we use exprtk, via `set_source_files_properties(foo.cpp PROPERTIES COMPILE_FLAGS XXXX)`. 

If `__cplusplus` could be used as a reliable macro to detect C++17, I would suggest @ArashPartow to use it to use the `[[fallthrough]]` attributes in the `exprtk_disable_fallthrough_begin` macro but... apparently the C++ standardization committee (for some reason?) didn't recomment updating the value of `__cplusplus` so there is no portable way to tell C++14 from C++17 (sigh)... See [this](https://stackoverflow.com/a/38456243/1631514).

I can't think of any immediate, simple way to allow exprtk sources to remain compatible with C++11/14/17 compilers and still use the newest `[[fallthrough]]` attribute. @ArashPartow could add a new level if `#if ...` to the definition of `exprtk_disable_fallthrough_begin`  checking for some arbitrary, yet-to-be-defined macro that means that C++17 is enabled; but that's a decision of the exprtk project, perhaps there is no much to gain with this idea (?).

So... I think we'll just merge the PR as is, ok? :-)

jolting:
I think it should be like this:
```
#if __has_cpp_attribute(fallthrough)
#define MRPT_FALLTHROUGH [[fallthrough]]
#elif __has_cpp_attribute(clang::fallthrough)
#define MRPT_FALLTHROUGH [[clang::fallthrough]]
#else
#define MRPT_FALLTHROUGH
#endif
```
-https://infektor.net/posts/2017-01-19-using-cpp17-attributes-today.html

jlblancoc:
Love it, thanks! Will try it (for this and other attributes) in mrpt2.
So... we will have our first macros to be removed when C++17 become the minimum required version of the language... endless story!

ArashPartow:
@jolting

That's an excellent suggestion, and I will indeed look to incorporating it for switches where there is one or more (but not all)  cases that require a fall-through.

-----

That being said, the current situation for ExprTk _(and code I write in general)_, is that either a switch statement breaks on all cases and has a default, or a switch falls-through on **all** cases. There would never be a situation, at least in ExprTk, for mixed modes of cases (break on some, fall through on others) in the same switch statement.

ArashPartow:
@jolting, @jlblancoc 

A quick note, according to the following answer, __has_cpp_attribute is a recent thing in at least GCC:

https://stackoverflow.com/a/30403394

Perhaps there may be a need to do an if-def on ___has_cpp_attribute_ itself, if the lower bound on GCC versions is still 4.3/4.6.

jolting:
Just wrap it in `#ifdef __has_cpp_attribute`

See the example here.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0061r0.html

PR #583, opened 2017-08-26
New GUI app to visualize and edit robot-made maps 

LisGein:
## Changed app:

* robot-map-gui

## PR Description
 Added new functionality:
- multiselection poses;
- deleting poses;
- Undo/Redo;
- Saving changes. 

Also, added the short description of classes.

---

I acknowledge to have:

(Notify: @MRPT/owners )


libs/obs/src/CSimpleMap.cpp
```diff
@@ -233,6 +233,26 @@ void CSimpleMap::insert(
 	insert(CPose3DPDF::Ptr(CPose3DPDF::createFrom2D(*in_posePDF)), in_SF);
 }
 
+void CSimpleMap::insertToPos(
+	int index, const CPose3DPDF::Ptr& in_posePDF,
+	const CSensoryFrame::Ptr& in_SF)
+{
+	int size = m_posesObsPairs.size();
```
jlblancoc: @LisGein : I know this is "unlikely" to happen in this case but... `int` typically has 32bit size on 32bit builds, while `.size()` returns `size_t` (or `std::size_t`) which is typically 64bit, so comparing them makes me nervous...  (in some extreme cases, like when comparing against `std::string::npos`, I have seen these errors leading to crashes!!).

The best solution would be use use `size_t` for the type of the incoming `index` parameter. 
Also, please, add a brief Doxygen documentation to the declaration of this method in the .h

Also: should the ">" in `(index > size)` be a ">=" ?? (Not sure, just asking... it depends on the API of `std::deque::insert()`)

(You can just add a new commit to your fork and it'll automatically added to this PR)

libs/gui/src/CAboutBoxBase.cpp
```diff
@@ -0,0 +1,89 @@
...
+		str += EIGEN_WORLD_VERSION + ".";
+		str += EIGEN_MAJOR_VERSION + ".";
+		str += EIGEN_MINOR_VERSION + "\n";
+
+		str += guiLibName + " version:      " + guiLibVersion;
+#if defined(__WXMSW__)
+		str += "-Windows";
+#elif defined(__UNIX__)
+		str += "-Linux";
+#endif
```
jlblancoc: Should we also add a Qt version line here?

jlblancoc:
@LisGein I added a pair of comments on minor issues. 
Also, please update the `AUTHORS` file to add yourself.

libs/gui/src/error_box.cpp
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+void GUI_IMPEXP mrpt::gui::showErrorMessage(const std::string& str)
+{
+#if MRPT_HAS_Qt5
+	QErrorMessage msg;
+	msg.showMessage(QString::fromStdString(str));
+	msg.exec();
+#else
```
jlblancoc: This should also check for wx and call wxMessageBox()...

libs/gui/src/CAboutBoxBase.cpp
LisGein: I use "guiLibName" for this, which defined for Qt in the constructor CAboutBoxQt and for wx in the  CAboutBox::OnInit.

jlblancoc:
@LisGein I detected these important (!!) warnings: 

```
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp: In member function ‘std::__cxx11::string CAboutBoxBase::information(const string&, const string&) const’:
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:63:32: warning: offset outside bounds of constant string
   str += EIGEN_WORLD_VERSION + ".";
                                ^~~
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:64:32: warning: offset outside bounds of constant string
   str += EIGEN_MAJOR_VERSION + ".";
                                ^~~
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:65:32: warning: offset outside bounds of constant string
   str += EIGEN_MINOR_VERSION + "\n";
                                ^~~~
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:63:32: warning: array subscript is above array bounds [-Warray-bounds]
   str += EIGEN_WORLD_VERSION + ".";
                                ^~~
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:63:32: warning: array subscript is above array bounds [-Warray-bounds]
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:65:32: warning: array subscript is above array bounds [-Warray-bounds]
   str += EIGEN_MINOR_VERSION + "\n";
                                ^~~~
/home/jlblanco/code/mrpt.lis/libs/gui/src/CAboutBoxBase.cpp:65:32: warning: array subscript is above array bounds [-Warray-bounds]
```

since `EIGEN_XXX` are numbers, not string literals. You could use `std::stringstream ss; ss <<  EIGEN_MINOR_VERSION << "." << ...;` or `str+=mrpt::format("%u.%u.%u",EIGEN_XXX,...);`.

jlblancoc:
A couple of warnings you may want to fix: 

```
/home/jlblanco/code/mrpt.lis/apps/robot-map-gui/CUndoManager.cpp:30:6: warning: unused variable ‘redoSize’ [-Wunused-variable]
  int redoSize = m_redo.size();
      ^~~~~~~~
/home/jlblanco/code/mrpt.lis/apps/robot-map-gui/CUndoManager.cpp: In member function ‘UserAction CUndoManager::undoAction()’:
/home/jlblanco/code/mrpt.lis/apps/robot-map-gui/CUndoManager.cpp:51:6: warning: unused variable ‘undoSize’ [-Wunused-variable]
  int undoSize = m_undo.size();

/home/jlblanco/code/mrpt.lis/apps/robot-map-gui/gui/observationTree/CRootNode.cpp:29:6: warning: unused variable ‘size’ [-Wunused-variable]
  int size = m_posesNode.size();
```

And regarding editing robot poses, some remarks: 
* When one clicks on a robot pose, it would be informative to know which "position" (pair: pose + observations) it is in the tree view of observations. Possible solutions (you may have better ideas) are: 
  * showing the 0-based index of the clicked pose... somewhere? 
  * highlighting the corresponding element in the tree-view?
* Since you insert an "XYZ corner" for each pose, you could (probably via another optional click button in the toolbar) show a text label for each corner with its index in the 3D view. It's easy since in MRPT every CSetOfObjects (actually, any CRenderizable class) has a `setName()` and `enableShowName()` which will minimize code changes in your app.
* BTW: in the tree of observations, now I realize we can't see their indices within the simplemap, right? Could we render them, instead of as it's now "Position", like: 
```
mrpt::system::TTimeStamp t = INVALID_TIMESTAMP;
// ... get t from the first CSensoryFrame entry, if not empty.
const std::string label = mrpt::format("[%u] Position (Timestamp: %s"), index, t != INVALID_TIMESTAMP ? mrpt::system::dateTimeLocalToString(t).c_str() : "Unavailable" );
```
* What about editing the *heading* (orientation) of a pose? Perhaps allowing changing a pose via textual edit boxes (for example, via a contextual menu on the pose entry in the tree? or anything else which is easy to implement in Qt?) will be more than enough.
* Removing an entry (pair: pose+obs) from the simplemap might also be really useful, and should be quite straightforward to implement...


These minor, minor details aside... the app looks great! 
Fantastic contribution, thanks!! :+1: 

I'll wait to hear your opinion on these last comments before merging. 
Cheers!

jlblancoc:
Hi @LisGein !

I can compile mrpt-gui, your new app, etc. correctly in my machine... but Travis keeps complaining about missing `#include`s (!). I've just added [some verbose output](https://github.com/mrpt-ros-pkg/mrpt_navigation/pull/83/) to see if that helps us to identify the problem... If you have a spare while to take a look at the next travis log and have a happy idea of what's wrong, it'll be welcome!

jolting:
That file isn't in trusty.
https://packages.ubuntu.com/trusty/amd64/qtbase5-dev/filelist

Only since 5.4
http://doc.qt.io/qt-5/qopenglwidget.html

Trusty uses 5.2

jlblancoc:
Absolutely right!

PR #616, opened 2017-11-21
Simplify arguments used in prepareFastDrawSample

jolting:
This fixes quite a few static casts and replaces them with
lambda functions, which should be more flexible.

Now the only argument prepareFastDrawSample passes into the callback
is size_t i. The rest of the arguments can be bound to the function.

Also removed some unnecessary includes of CParticleFilterCapable.h.

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/apps/ro-localization/CPosePDFParticlesExtended.cpp
+++ b/apps/ro-localization/CPosePDFParticlesExtended.cpp
@@ -39,19 +39,17 @@ namespace mrpt
 {
 namespace poses
 {
-
 class MyStandardProposal
 {
-public:
+   public:
 	static constexpr bool DoesResampling = true;
 };
 
 class MyAuxiliaryPfOptimal
 {
-public:
+   public:
 	static constexpr bool DoesResampling = false;
 };
-
 
 IMPLEMENTS_SERIALIZABLE(CPosePDFParticlesExtended, CPosePDF, mrpt::poses)
 
@@ -102,8 +100,7 @@ void CPosePDFParticlesExtended::copyFrom(const CPosePDF& o)
 		m_particles.clear();
 		m_particles.resize(M);
 
-		for (itDest = m_particles.begin(),
-			partsIt = parts.begin();
+		for (itDest = m_particles.begin(), partsIt = parts.begin();
 			 itDest != m_particles.end(); itDest++, partsIt++)
 		{
 			itDest->log_w = 0;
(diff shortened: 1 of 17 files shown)


apps/ro-localization/CPosePDFParticlesExtended.cpp
```diff
@@ -39,20 +39,18 @@ namespace mrpt
 {
 namespace poses
 {
-
 class MyStandardProposal
 {
-public:
+   public:
 	static constexpr bool DoesResampling = true;
```
jlblancoc: Just curious: does a `static (integral type) foo = value;`  really change at all by adding `constexpr`? Hmmm...

jlblancoc:
love it! 👍

apps/ro-localization/CPosePDFParticlesExtended.cpp
jolting: Slight performance improvement.
if( someconditional && constexprisalwaysfalse)
{
   this is optimized out.
}

https://github.com/MRPT/mrpt/blob/8dc65be987aa47e90c991d745843509fedab6da8/libs/base/include/mrpt/bayes/CParticleFilter_impl.h#L65

Also good for meta programming.

jlblancoc: Alright. I think I was (incorrectly?) relying (and hoping) that compilers were able to automatically optimize out `const static` values as if they were `constexpr`. 
Anyhow, adding it makes clear our intentions to **any** compiler, so it's good!

jolting: Regular const can always be reinitialized to something else in the constructor, so in order to make that optimization you would have to know something about all the constructors. I think that for static member variables that is not a common optimization. I could be wrong. 

For thing in a local scope I would imagine it is more common to make that optimization because it's easier to know statically the const can only ever be one value. Constexpr should always be the safe bet though,  even for local variables.

jolting: By the way, Here's a good candidate for constexpr
https://github.com/MRPT/mrpt/blob/4511e27a6e53163cca287ee8cc7ac2b686f8efc0/libs/base/include/mrpt/poses/CPoint3D.h#L106

jlblancoc: Sure, thanks for the pointer! 👍 All metaprogramming stuff used in "traits" structs should be ported. 
I'll first finish, I think, the *huge* refactoring effort under way in the other branch, since it will create many conflicts otherwise. 
In fact, I should have to decide on integrating these changes from you on PFs into the other branch soon, before renaming/moving any mrpt-bayes file (!).

PR #636, opened 2018-01-24
fixed deadlock in CMetricMapBuilderRBPF::clear()

ceccocats:
## Changed apps/libraries

* modified lib mrpt::slam::CMetricMapBuilderRBPF

## PR Description
when calling  **CMetricMapBuilderRBPF::initialize( [...] )** the _critZoneChangingMap_ lock is taken, then the function call **this->clear()** and the same lock is taken twice.
This PR simply take the lock after calling the clear function. 

---

I acknowledge to have:

(Notify: @MRPT/owners )


jlblancoc:
👍 Thanks! 

Another option would be to replace std::mutex with std::recursive_mutex, just in case this (may) happen somewhere else...

ceccocats:
Nice, i have never used this types of locks

PR #642, opened 2018-02-06
Use codecov

jolting:
LICENSE is BSD
Based on this repo:
https://github.com/RWTH-HPC/CMake-codecov

Note this changes the behavior of make test. The old behavior is make test_legacy.

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/travis/travis_main.sh
+++ b/travis/travis_main.sh
@@ -93,6 +93,7 @@ function test ()
   cmake $MRPT_DIR \
     -DBUILD_APPLICATIONS=FALSE \
     -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
+    -DENABLE_COVERAGE=On \
     $EXTRA_CMAKE_ARGS
 
   # Remove gdb use for coverage test reports.
@@ -102,6 +103,7 @@ function test ()
 #  else
     make test
 #  fi
+  bash <(curl -s https://codecov.io/bash)
 
   cd $MRPT_DIR
 }
--- a/travis/travis_main.sh
+++ b/travis/travis_main.sh
@@ -101,7 +101,7 @@ function test ()
 #  if command_exists gdb ; then
 #    make test_gdb
 #  else
-    make test
+    make tests_build_all test
 #  fi
   bash <(curl -s https://codecov.io/bash)
 
--- a/travis/travis_main.sh
+++ b/travis/travis_main.sh
@@ -104,7 +104,6 @@ function test ()
     make tests_build_all
     make test
     make gcov
-    make lcov
 #  fi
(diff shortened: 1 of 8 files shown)


jolting:
Here is what it looks like:
https://codecov.io/gh/jolting/mrpt/branch/master

jlblancoc:
Looks great! 👍 

I've just tried this new setup under CMake+MSVC, and it complains code coverage is not supported with MSVC2017 (?). Anyway, we can live with that. 

The point is, the target "test" is now "test_legacy", and I can see a new CMake target "RUN_TESTS", which tried to run all examples without building them first (!). Perhaps the behavior for Windows is different than with Makefile generators? Will check it out later... 

In any case, I'm unconfortable with this line: 
https://github.com/MRPT/mrpt/pull/642/files#diff-7728197db21ae89029c6a9aff43f1e81R104

Unless dependencies are well defined, `make tests_build_all test` might try to run the tests without building them first? Also, if deps are correct, why not directly skipping the target tests_build_all? I think I'm missing something :-)

jolting:
ctest takes over the `make test` rule, but doesn't build the targets. Perhaps ctest `make test` rule can have target dependencies. I justs couldn't figure it out.

It's based on this:
https://github.com/RWTH-HPC/CMake-codecov/blob/master/.travis.yml

Maybe it's better just to copy and paste this:
```
  - cmake .. -DENABLE_COVERAGE=On && make && make test
  - make gcov
  - make lcov
```

travis/travis_main.sh
```diff
@@ -93,15 +93,20 @@ function test ()
...
     -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
+    -DENABLE_COVERAGE=On \
     $EXTRA_CMAKE_ARGS
 
   # Remove gdb use for coverage test reports.
   # Use `test_gdb` to show stack traces of failing unit tests.
 #  if command_exists gdb ; then
 #    make test_gdb
 #  else
+    make tests_build_all
```
jolting: It does each step independently now.

jolting:
Hmm added the `make lcov`, but it didn't work. Maybe lcov isn't installed in the container.

jolting:
The issue:
https://gitlab.kitware.com/cmake/cmake/issues/8774

jlblancoc:
I think I don't understand what do [the percentage figures here](https://codecov.io/gh/jolting/mrpt/branch/master) actually mean? 
Is really 80% the actual code coverage of *all* code under `libs`, when running all unit tests? I wish it was, but I think the number was pretty smaller when I did some cov tests in the past (?).

PR #649, opened 2018-02-11
Add option for extra libs for examples

jolting:
This builds now
---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/cmakemodules/DefineExamples.cmake
+++ b/cmakemodules/DefineExamples.cmake
@@ -331,14 +331,21 @@ IF(BUILD_EXAMPLES)
 	# === OPENNI2 examples ===
 	IF (MRPT_HAS_OPENNI2)
 		SET(LIST_EXAMPLES_IN_THIS_DIR
-			hwdrivers_openni2_rgbd_demo
-			hwdrivers_openni2_proximity_demo
 			hwdrivers_openni2_driver_demo
 			hwdrivers_openni2_2d_icp_slam
-			hwdrivers_openni2_to_rawlog)
-
+			)
 		SET(CMAKE_EXAMPLE_DEPS mrpt-hwdrivers mrpt-gui mrpt-opengl mrpt-maps)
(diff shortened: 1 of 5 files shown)


jolting:
I'm not sure if this is the best approach. I was also wondering if there is a way to add the generated cmake files to .gitignore

jlblancoc:
> I was also wondering if there is a way to add the generated cmake files to .gitignore

I know it seems redundant to include in git all CMakeLists.txt files for examples, when they can be dynamically generated, but it's done like that for two main reasons: (1) the most important is allowing users to directly copy/download/whatever an example subdirectory and build it **as is** with cmake after installing MRPT via apt-get or Windows installers; (2) I'm less sure of this one, but I don't feel comfortable modifying the SOURCE_DIR of CMake... I think that we should support building as if the the source code directories were read-only, just in case that's actually the case somewhere (?).

> I'm not sure if this is the best approach. 

Hmm... I see the problem. Probably it's better to add OpenNI2 libraries as CMake PUBLIC link dependencies. I'll try to test that approach, but won't be before late afternoon (CET time!), so let me know if you tried it first.

Cheers!

jolting:
Perhaps a custom target with a PUBLIC link dependency. I'd hate to taint one of the real library targets.

jlblancoc:
Let's keep this one open after #651, just to see if I can find an alternative "cleaner" solution with the PUBLIC interface way...

jolting:
Rebased for easier comparison

jlblancoc:
Thanks! 

I think the easier is to move these three examples to `doc/mrpt_*_examples/` together with other "special" examples, then modify their CMake files to do a pkg_config-search for openni2 in a proper way... 

Will try to do it later tonight.

jlblancoc:
Closed after: https://github.com/MRPT/mrpt/commit/8f2bbd99eb2d758c0f2103e65f5c1a91e89403d9

PR #657, opened 2018-02-18
Convert some typedefs to using

jolting:
I didn't get all typedefs, but it's a start.

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/bayes/include/mrpt/bayes/CKalmanFilterCapable.h
+++ b/libs/bayes/include/mrpt/bayes/CKalmanFilterCapable.h
@@ -240,15 +240,15 @@ class CKalmanFilterCapable : public mrpt::system::COutputLogger
 	}
 	inline bool isMapEmpty() const { return detail::isMapEmpty(*this); }
 	/** The numeric type used in the Kalman Filter (default=double) */
-	typedef KFTYPE kftype;
+	using kftype = KFTYPE;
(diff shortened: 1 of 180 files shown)


jlblancoc:
wow, great! I manually changed a few of them, but clearly not enough... 

cc: #66

jlblancoc:
Also, many complex typedefs and cumbersome for() loops **could** be converted to `auto` and ranges. I already changed a few as I stepped into them, but nothing systematic yet. Also, in some cases it might be informative to leave the real types instead of `auto` (mainly, in heavily template-based code where it's easy to lose grip).

jolting:
Most of these were done using a sed replacement.

4935 remaining typedefs, but that includes some of the otherlibs, which I avoided.
22.5% of remaining typedefs are in exprtk.hpp

Also my sed replacement trick missed all the multi-line typedef and the function typedefs.

jlblancoc:
> Also my sed replacement trick missed all the multi-line typedef and the function typedefs.

I'm annoyed by how hard it seems to be to run regex replacements when dealing with multilines... (sigh) 
After much wasted time trying it with command line tools and with Atom, the best similar thing I found (though, not 100% regex compatible) is Notepad++... a Windows GUI tool :-( I feel like I must be missing "The Easy Way" (!?)

PR #670, opened 2018-03-03
Convert more exception macros into inlined function

jolting:
This removes a bunch of shadow variable warnings.


jolting:
This gets rid of most of the shadow variable warnings.

PR #672, opened 2018-03-03
Refactor alignment in PointsMap

jolting:
This might need some unit tests. mrpt-performance doesn't seem to show degradation.

### DIFF
--- a/libs/maps/include/mrpt/maps/CPointsMap.h
+++ b/libs/maps/include/mrpt/maps/CPointsMap.h
@@ -12,6 +12,7 @@
 #include <mrpt/serialization/CSerializable.h>
 #include <mrpt/config/CLoadableOptions.h>
 #include <mrpt/core/safe_pointers.h>
+#include <mrpt/core/aligned_std_vector.h>
 #include <mrpt/math/KDTreeCapable.h>
 #include <mrpt/obs/CSinCosLookUpTableFor2DScans.h>
 #include <mrpt/math/lightweight_geom_data.h>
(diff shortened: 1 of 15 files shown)


libs/maps/include/mrpt/maps/CPointsMap.h
```diff
@@ -1090,7 +1091,7 @@ class CPointsMap : public CMetricMap,
 
    protected:
 	/** The point coordinates */
-	std::vector<float> m_x, m_y, m_z;
+	mrpt::aligned_std_vector<float> m_x, m_y, m_z;
```
jlblancoc: Oh! Why I didn't thought of this?
Perhaps it wasn't needed to make all containers aligned (weight, color,...) but at least X,Y,Z will be useful **for sure**.  :+1:

libs/serialization/src/CArchive.cpp
```diff
@@ -236,6 +237,11 @@ CArchive& mrpt::serialization::operator<<(
 {
 	return detail::writeStdVectorToStream(s, a);
 }
+CArchive& mrpt::serialization::operator<<(
+	CArchive& s, const mrpt::aligned_std_vector<float>& a)
+{
+	return detail::writeStdVectorToStream(s, a);
```
jlblancoc: Do you know if there is any syntax to let the compiler know that a pointer is aligned? something like `align(16) float*`? I ask because using an aligned_std_vector<> will not give that information, in spite of the aligned allocator. Just thinking ahead...

libs/maps/include/mrpt/maps/CPointsMap.h
jolting: Yeah. Might have been overkill for color. Sometimes color has useful information for processing and the algorithims are worth accelerating.

libs/serialization/src/CArchive.cpp
jolting: Not that i know of. You can build something like gsl not_null<T>.

libs/maps/src/maps/CPointsMap.cpp
```diff
@@ -462,6 +424,23 @@ void CPointsMap::determineMatching2D(
 	local_y_max =
 		max(max(temp_nums[0], temp_nums[1]), max(temp_nums[2], temp_nums[3]));
 
+	for (size_t k = 0; k < nLocalPoints % 4; k++)
```
jolting: It seems like it would be easier to handle the last 3 separately instead of padding with zeros.

jlblancoc: Thanks for the patch! Yes, it's clearer like that.

PR #673, opened 2018-03-04
Organize Travis builds into stages

jolting:


jlblancoc:
Great idea! 
It seems there is plenty of room (time) left in stage 1, so we could try building some more libs [up the tree of dependencies](http://mrpt.ual.es/reference/master/), for example, mrpt-slam, to leave more free time in next stages.

jlblancoc:
Ok, I saw you commit... we are really in synch XD

jolting:
I can't see a limit on dockerhub, so it made sense.

jlblancoc:
It works like a charm! 🥇

jolting:
Hopefully, travis builds will be reliable now.

PR #680, opened 2018-03-07
More constexpr in light geom

jolting:


jolting:
Hmmm. I think this should be supported in c++14. Looks like the version of gcc we are using disagrees.

jolting:
C++14 should support it.

Here's one bug:
https://stackoverflow.com/questions/36489369/constexpr-non-static-member-function-with-non-constexpr-constructor-gcc-clang-d

Here's another bug:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67371

Very disappointing.

jolting:
Ubuntu 18.04 is coming in April. MRPT 2.0 could target that compiler. :sweat_smile:

jlblancoc:
Yes, it's frustrating... I've been using g++-7 in u16.04 for a long time (I can confirm your patch does build with these settings), so I would be happy of raising the requisites so we don't let gcc-5 drags us!

I've just [pushed](https://github.com/MRPT/mrpt/pull/680/commits/3f6b171f6bddfc4f65f962c5fb932134dba1089f) a patch for the README. 

But before going on with this, I think we should also clear two dependencies: 
* Travis: Could you please update the docker imgs so the u16.04 ones include gcc7? (see the commands in the patch to the readme above).
* PPA builds (I'll see if I can add the gcc-7 as a depending PPA, it should work).

jolting:
Docker images building.

jlblancoc:
Thanks! I restarted travis jobs, let's see...

jlblancoc:
@jolting It [still finds](https://travis-ci.org/MRPT/mrpt/jobs/350212870) gcc-5 in the docker img in travis (?). Perhaps related to caching (CMakeCache.txt) into the container? 

I already pushed the changes to build against gcc-7 in the PPA, let's see if that one works on the next automatic PPA update...

jolting:
What about CXX=g++-7 in the env?

jlblancoc:
> What about CXX=g++-7 in the env?

Done now (in master). Let's wait for [this](https://launchpad.net/~joseluisblancoc/+archive/ubuntu/mrpt/+packages) to build to see if no error happens, then we can merge.

PR #695, opened 2018-03-21
New pimpl pattern

jolting:


jolting:
@jlblancoc Do you know if this work on windows?

jlblancoc:
Yes, it works like a charm! 👍 

Before merging: do you think it's worth changing the existing use cases of `core/pimpl.h` to this new scheme? Probably not...

jolting:
I'll do that later. The hard part is to the need for a deep copy on those pointers to make those classes copyable.

jlblancoc:
spimpl is a great discovery! 👍 
I guess propagate_const is no longer needed, right?

jolting:
Yeah. Wasn't sure if it was keeping around. Non pimple member pointers should probably use it.

jlblancoc:
👍 merged!

PR #708, opened 2018-04-06
Externalpackage jpeg

jolting:


jlblancoc:
I can confirm the linking warning has gone after /MD 👍

Issue #703, opened 2018-03-31
Handle LGPL more cleaning


jolting:
LGPL is stricter than BSD, so it simply can't be relicensed as BSD code. It should be clear what is LGPL derived code. Ideally, LGPL code should be in its own library(.so) when linking to BSD code.

Based on explicit mentions of LGPL in MRPT code here are my suggestions.
- [ ] libs/vision/src/faster/*
Faster can be in libmrpt-vision-faster.so

libmrpt-vision-faster.so would be LGPL.

- [ ] libs/vision/src/ba_full.cpp and libs/vision/include/mrpt/vision/bundle_adjustment.h
libmrpt-vision-ba.so
libmrpt-vision-ba.so would be LGPL.

- [ ] libs/system/include/mrpt/system/memory.h
Rewrite is_aligned from scratch or copy a BSD implementation.

Still undecided
- [ ] libs/opengl/src/gl_utils.cpp
- [ ] libs/opengl/include/mrpt/opengl/CText3D.h
- [ ] libs/opengl/include/mrpt/opengl/gl_utils.h
- [ ] libs/img/src/CImage_SSE2.cpp

jlblancoc:
@jolting I think your recent branch would solve all the pending items here, right?

jolting:
libs/vision/src/ba_full.cpp and libs/vision/include/mrpt/vision/bundle_adjustment.h actually needs to be solved.

Issue #120, opened 2015-01-07
Clean-up of mrpt-vision API


jlblancoc:
To-do list: 
- [x] Upgrade IplImage -> cv::Mat (See #710)
- [x] Re-think features tracker, if needed.
- [x] Remove un-needed classes.
- [x] Improve docs in lib doxygen page.
- [x] Go thru all the `MRPT_TODO`'s in this library and resolve them.
- [x] Delete all old SIFT implementations
- [x] Remove CFeatureList

jlblancoc:
Update on this one: 
I think `CFeatureList` should be entirely removed, leaving only `TSimpleFeatureList` for feature points, and a new alternative to `CFeatureList` for storing **descriptors**. 

Part of this was done in #889 , but the alternative to `CFeatureList` is left as future work since it's not urgent for me right now. 

Note that `CFeatureList`  is used in some past-years GSoC GUI apps, so refactoring would imply some amount of work.

IMO, this ticket would be a good GSoC 2019 project, BTW...

PR #711, opened 2018-04-09
Use libcvd as an external package

jolting:
This helps solve #703.


jlblancoc:
Thanks for taking the Licensing issues seriously 👍 
This is a fat addition, but as long as all TooN & CVD headers remain in the implementation side (no exposure to the user), as it's the case, it's fine. We will now have at hand the power of the entire CVD, which will also help (someday...) with #120 as well. I manually checked it builds cleanly on MSVC too.

I'm glad of incorporating code by Georg Klein and friends.

jolting:
>This is a fat addition, but as long as all TooN
I don't think we need TooN

> CVD headers remain in the implementation side (no exposure to the user),
ExternalPackage could easily point to an external repo instead of otherlibs, but that would require conditionally compiling the MRPT classes if you don't have CVD.

I added a few minor patches to their cmake file to get it to compile on Windows/Linux.

PR #715, opened 2018-04-18
Support shared_ptr<T> class types in serialization and typemeta modules

jlblancoc:
That's it. Refer to the modified examples and docs.

`shared_ptr`'s can now be safely (de)serialized, not only for CSerializable classes, but for POD or custom user's classes.


jolting:
LGTM

PR #720, opened 2018-04-22
Output tests to bin dir

jolting:
Fixes appveyor

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,7 +5,7 @@ INCLUDE(../cmakemodules/AssureCMakeRootFile.cmake) # Avoid user mistake in CMake
 # ----------------------------------------------------------------------------
 PROJECT(tests)
 
-SET(EXECUTABLE_OUTPUT_PATH "${${PROJECT_NAME}_BINARY_DIR}")
+SET(EXECUTABLE_OUTPUT_PATH "${MRPT_BINARY_DIR}/bin")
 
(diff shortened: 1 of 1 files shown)


jlblancoc:
I just saw it, yep... 
There was a "delayed load" command in hwdrivers related to libpcap, but it should only be enabled if pcap is found: https://github.com/MRPT/mrpt/blob/master/libs/hwdrivers/CMakeLists.txt#L193-L201
Perhaps a pragma link...

jlblancoc:
https://github.com/MRPT/mrpt/blob/master/libs/hwdrivers/src/CIMUXSens_MT4.cpp#L316-L320

https://github.com/MRPT/mrpt/blob/master/libs/hwdrivers/src/CIMUXSens.cpp#L45-L48

jlblancoc:
Right. Then, my only suspects are the two pragmas in the comment above: 
WinUsb.dll
Setup.dll

Perhaps WinUSB.dll would make sense to be something a virtual container may be stripped off (?).

jolting:
What about assimp?

jlblancoc:
Try please adding a commit disabling both XSens 3 & 4... if that fixes the issue we'll have a positive finding.

jolting:
hwdrivers maps nav obs opengl slam and topography depend on assimp. The rest of the tests don't. I think thats why hwdriver is the first one to fail. assimp dll is missing when those tests are running.

Issue #725, opened 2018-04-23
Build error on 16.04 "m_partitioner.addMapFrame"


cressell:
Hi,

I'm getting an error when trying to build the master branch on ubuntu 16.04 (jetson tx2). I have updated gcc+.

`error: no matching function for call to ‘mrpt::slam::CIncrementalMapPartitioner::addMapFrame(mrpt::obs::CSensoryFrame&, std::__shared_ptr_access<mrpt::poses::CPosePDF, (__gnu_cxx::_Lock_policy)2, false, false>::element_type&)’
   m_partitioner.addMapFrame(sf, *posePDF);`

Above is the error I'm getting. Any clues?

Regards,
Cressell

jlblancoc:
This should be fixed after #726 , if not let us know... thanks!

Issue #723, opened 2018-04-22
Bug in CLoopCloserERD


MaxGsomGsom:
I have spent a lot of time to find why my configuration of graph slam don't work. And it seems that there should be ``if (!it->second)``.

https://github.com/MRPT/mrpt/blob/50111a08efeb0744d5e97658b30f9131c5fef5f3/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h#L2515-L2527

jolting:
CC: @bergercookie

bergercookie:
Yeah, this seems to be a bug.
I'm surprised this is there because it's quite big for me not to notice it.

Anyway, feel free to make a PR for this, otherwise I can patch it in the couple of days.

MaxGsomGsom:
Also it does not compile:
https://github.com/MRPT/mrpt/blob/50111a08efeb0744d5e97658b30f9131c5fef5f3/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h#L2539-L2548

Because
https://github.com/MRPT/mrpt/blob/50111a08efeb0744d5e97658b30f9131c5fef5f3/libs/slam/include/mrpt/slam/CIncrementalMapPartitioner.h#L137-L139

jlblancoc:
Just fixed with 630ad757efc9873f64c735f36e6afe977c2a056f

MaxGsomGsom:
I have tried to compile again but it shows error:
``/home/max/Desktop/mrpt/libs/slam/include/mrpt/slam/CIncrementalMapPartitioner.h:137:11: note:   no known conversion for argument 2 from ‘std::__shared_ptr_access<mrpt::poses::CPosePDF, (__gnu_cxx::_Lock_policy)2, false, false>::element_type {aka mrpt::poses::CPosePDF}’ to ‘const mrpt::poses::CPose3DPDF&’``

jolting:
That might be an explicit conversion. Try `mrpt::poses::CPose3DPDF(*posePDF)`

jlblancoc:
Should be fixed after #726, if not please reopen...

PR #726, opened 2018-04-23
mrpt-graphslam: Fix compilation error in CLoopCloser

bergercookie:
## Changed apps/libraries

mrpt-graphslam

## PR Description

This should most likely deal with https://github.com/MRPT/mrpt/issues/725, https://github.com/MRPT/mrpt/issues/723

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
+++ b/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
@@ -2537,14 +2537,15 @@ void CLoopCloserERD<GRAPH_T>::updateMapPartitions(
 		}
 
 		// pose
-		pose_t curr_pose = search->second;
-		mrpt::poses::CPosePDF::Ptr posePDF(new constraint_t(curr_pose));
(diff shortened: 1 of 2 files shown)


jolting:
I'll glad createFrom2D exists, but can that be changed to an explicit conversion constructor?

I would deprecate or remove that way of conversion because it is inconsistent with the c++ standard.

jolting:
http://en.cppreference.com/w/cpp/language/using_declaration#Inheriting_constructors

jlblancoc:
I can't remember why the `createFrom2D()` was introduced in the first term, but the `using B::B;` rule is huge! I'll check it out... :+1:

jlblancoc:
ok, I've already checked it: this particular case cannot be mapped to a regular explicit ctor, becase it's actually a class factory with the type to be constructed determined at runtime. 
That is: it constructs an object of type `CPose3DPDF{a,b,c}` if fed with `CPosePDF{a,b,c}`. After a little thinking, I can't see any way of getting rid of the static factory method...

PR #731, opened 2018-04-25
Simplifications in CLoopCloser_impl.h 

bergercookie:
## Changed apps/libraries

mrpt-graphlam

## PR Description

Apply correction proposed in https://github.com/MRPT/mrpt/issues/723.

@jlblancoc even after this fix it seems that running graphslam-engine with CLoopCloser doesn't run correctly. CLoopCloser utilises the CIncrementalMapPartitioner for partitioning the graph under construction (so that potential LCs are detected and placed in the same partitions). However, when running it, every node is placed on a partition of its own!

Is there any change in the CIncrementalMapPartitioner that I am not aware of (e.g., some change in the .ini parameters or the values they accept)? 

Command that I ran to test this:

`./bin/graphslam-engine --2d -r ~/mrpt/share/mrpt/datasets/graphslam-engine-demos/action_observations_map/simul.rawlog -i ~/mrpt/share/mrpt/config_files/graphslam-engine/odometry_2DRangeScans_LC.ini -e CLoopCloserERD`

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
+++ b/libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
@@ -44,9 +44,8 @@ CLoopCloserERD<GRAPH_T>::~CLoopCloserERD()
 
 	// release memory of m_node_optimal_paths map.
 	MRPT_LOG_DEBUG_STREAM("Releasing memory of m_node_optimal_paths map...");
-	for (typename std::map<mrpt::graphs::TNodeID, path_t*>::iterator it =
-			 m_node_optimal_paths.begin();
-		 it != m_node_optimal_paths.end(); ++it)
+	for (auto it = m_node_optimal_paths.begin();
+			 it != m_node_optimal_paths.end(); ++it)
 	{
 		delete it->second;
 	}
@@ -145,7 +144,6 @@ bool CLoopCloserERD<GRAPH_T>::updateState(
 			 this->m_just_inserted_lc)
 				? true
 				: false;
-
 		this->updateMapPartitions(
 			m_partitions_full_update,
 			/* is_first_time_node_reg = */ num_registered == 2);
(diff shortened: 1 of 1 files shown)


bergercookie:
**Update:** Just noticed that @jolting has already fixed the inversion of checking the laser scan.

libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
```diff
@@ -2515,9 +2504,7 @@ void CLoopCloserERD<GRAPH_T>::updateMapPartitions(
 	// TODO - Should always exist.
 	// for each one of the above nodes - add its position and correspoding
 	// laserScan to the partitioner object
-	for (typename nodes_to_scans2D_t::const_iterator it =
-			 nodes_to_scans.begin();
-		 it != nodes_to_scans.end(); ++it)
+	for (auto it = nodes_to_scans.begin(); it != nodes_to_scans.end(); ++it)
```
jolting: This can be replaced by a ranged based for.

libs/graphslam/include/mrpt/graphslam/ERD/CLoopCloserERD_impl.h
```diff
@@ -44,9 +44,8 @@ CLoopCloserERD<GRAPH_T>::~CLoopCloserERD()
 
 	// release memory of m_node_optimal_paths map.
 	MRPT_LOG_DEBUG_STREAM("Releasing memory of m_node_optimal_paths map...");
-	for (typename std::map<mrpt::graphs::TNodeID, path_t*>::iterator it =
-			 m_node_optimal_paths.begin();
-		 it != m_node_optimal_paths.end(); ++it)
+	for (auto it = m_node_optimal_paths.begin();
```
jolting: Another range-based for

jlblancoc:
@bergercookie : 

> LCs are detected and placed in the same partitions). However, when running it, every node is placed on a partition of its own! 
> Is there any change in the CIncrementalMapPartitioner that I am not aware of (e.g., some change in the .ini parameters or the values they accept)? 

Yes, the map partitioner class was recently slightly redesigned, but afterwards I've verified it works, so it may probably be that you need to set some of its new parameters (?). 

Please, see this diff for you to understand what exactly I changed in that class API: https://github.com/MRPT/mrpt/pull/716/files (click on "jump to..." CIncrementalMapPartitioner .h and CIncrementalMapPartitioner .cpp). 

From the error you describe, please check this: 
* that you set "simil_method": I should have updated your graphslam call, replacing the old equivalent parameter "useMapMatching" to the corresponding enum, but please double verify it.
* call getAdjacencyMatrix(), before or after updating the partitions, it doesn't matter, and let's see how does it look like. Plot it with `spy()` or as a colormap 2D image in matlab/octave if it's large to better visualize it.

jlblancoc:
See also an example of how to set the `maxDistForCorr` value. Set it to its old default (0.1 or 0.2 I think) and give it a try (that could be a reason for what you observe):
https://github.com/MRPT/mrpt/pull/716/commits/51a636f7a8d8f4c47b7c51a3a7ceb99d581313cd#diff-8640e773b3f2ca73ad48cac2e01427caL62

jolting:
Can we merge this one?

jlblancoc:
Yes, let's merge it and if @bergercookie finds new bugs/changes it would go into a new PR.

PR #733, opened 2018-04-26
Separate lgpl bundle adjustment code into it's own library and turn it off by default

jolting:


jolting:
This is the last thing I found for #703.
The anti-tivoization clause is going to be the biggest reason to avoid LGPLv3. This might make a few users a little more comfortable.

PR #734, opened 2018-04-26
Velodyne data packet publish

lasdasdas:
## Changed apps/libraries
hwdrivers::CVelodyneScanner class

## PR Description
Added the option to retrieve single data packets  when calling getNextObservation(). This option is disabled by default and setFramePublishing() is the member function used to set this option. setFramePublishing(true) makes getNextObservation() return full revolutions of points, and return single data packets when set to false. 
---
This has been tested on a VLP16, and on pcaps of HDL32 and VLP16 and compiles correctly with gcc

I acknowledge to have:
* Read the [`CONTRIBUTING`](https://github.com/MRPT/mrpt/blob/master/.github/CONTRIBUTING.md) page -> yes
* Updated [`doc/doxygen-pages/changeLog_doc.h`](https://github.com/MRPT/mrpt/blob/master/doc/doxygen-pages/changeLog_doc.h) to describe these changes (if applicable) -> no need, but a brief description was added to explain the member function. 
* Updated [`AUTHORS`](https://github.com/MRPT/mrpt/blob/master/AUTHORS) (if applicable) -> done

@MRPT/owners

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -162,5 +162,8 @@ the GitHub contributors page (https://github.com/MRPT/mrpt/graphs/contributors).
   - Driver for SICK TIM561/55x/56x via ethernet (mrpt::hwdrivers::CSICKTim561Eth)
   - A testing sample for SICK TIM55x/TIM56x series
 
+* David Serret
+  - Small changes in Velodyne drivers
+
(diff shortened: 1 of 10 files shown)


jolting:
Seems reasonable

Issue #581, opened 2017-08-22
Use C++17 attributes when available


jlblancoc:
See [this comment](https://github.com/MRPT/mrpt/pull/580#issuecomment-323892035).

Useful attributes (feel free of editing the list):
* [x] `[[maybe_unused]]`  <-- MRPT_UNUSED_PARAM
* [x] `[[fallthrough]]`
* [x] `[[deprecated]]`
* [ ] `[[nodiscard]]`
* [ ] `[[gnu::pure]]` and `[[gnu::const]]` (functions only)

jlblancoc:
Just added to the issue description a list of useful attributes to add.

jlblancoc:
Just added `[[gnu::pure]]` and `[[gnu::const]]`

PR #735, opened 2018-04-28
particle filters: use values vs. pointers; enable C++17 in MRPT

jlblancoc:
* CProbabilityParticle now has a "storage" template param.
* SE(2) / SE(2) particle filters now use value storage instead of smart pointers for robot poses, saving huge numbers of memory allocations.
* MRPT now requires C++17 to build. See minimum compilers in updated README.md

Pending before merging: 
* [x] Take a look at what's wrong with the python bindings. I fixed many error there but the remaining one, I can't localize the problem origin...


jolting:
Ok #581 is now a priority.

jolting:
Wait for the docker hub containers to be ready and then change the travis script.

jolting:
Looks like gtest is broken.
```
  c:\projects\mrpt\otherlibs\gtest-1.8.0\include\gtest\internal\gtest-port.h(996): error C2039: 'tr1': is not a member of 'std' [c:\projects\mrpt\build\otherlibs\gtest-1.8.0\mrptgtest.vcxproj]
```
Also the include of windows.h in mrpt\libs\gui\src\CDisplayWindow3D.cpp is causing problems.
Try `#define WIN32_LEAN_AND_MEAN` before the include.

jolting:
I bet defining `GTEST_USE_OWN_TR1_TUPLE` will fix one of those problems.

jolting:
I seem to recall an issue with the python bindings with c++17 on Xenial. What about just disabling the python bindings on 16.04?

jlblancoc:
The error was so intrincated I couldn't recognize what was the actual
problem... even after years of experience with templates 😠 I'll give it
another try after having had some sleep and let's decide then!

That's the only remaining error in my last night pr.

jlblancoc:
Please, don't merge yet. I'm fixing / avoiding the python issues. 
But the key is that the idea of converting the smart pointers inside particles to values **seems** to have decreased the execution speed of PF localization (!!). I want to double check what's going on...

jolting:
Just added liblas-c-dev to travis full-bionic. That should enable nearly everything in Bionic.

jolting:
>Please, don't merge yet. I'm fixing / avoiding the python issues.
But the key is that the idea of converting the smart pointers inside particles to values seems to have decreased the execution speed of PF localization (!!). I want to double check what's going on...

Probably a missing move somewhere.

jlblancoc:
False alarm: efficiency is virtually the same in master vs. this branch. I'll later compare both to 1.5, since it seems to me (may be a delusion) that the demos were faster. 

AnywaY: we can merge now. Python errors have been isolated by commenting out the problematic code fragments. Not a real fix but allows us moving forward.

jlblancoc:
> Probably a missing move somewhere.

I'll recheck the resampling functions anyway. But in the case of pf-localization, particles hold data as small as 3 double's, so moving or copying should be identical and should have a "negligible" cost.

jlblancoc:
@jolting can be also delete branch dev-simpler-pose-particles-bionic? I don't want to do it just in case you have WIP over there...

PR #740, opened 2018-05-01
Replace mapbox::variant with std::variant

jolting:


jlblancoc:
Good! Last night I was thinking about this... but thought it was optional, not variant, so didn't find it hehe

jolting:
I think variants might potentially solve the static build problem too.

jlblancoc:
Unfortunately, it seems to broke the examples: 
https://travis-ci.org/MRPT/mrpt/jobs/373676786#L552

Looks like `.match` was a mapbox's thing...

jolting:
Woops. That's std::visit now.

jlblancoc:
Thanks! Let me fix it, so I get used to these C++17 new stuff... ;-)

PR #741, opened 2018-05-02
use std::visit()

jlblancoc:
Fixes mapbox's `t.match()` ==> `std::visit()`.

PS: First test after setting mrpt/master as a Github "protected branch". Travis CI (and only Travis) being green is now a pre-requisite for merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Minimum compiler requisites:
     * gcc-7 or newer.
       * Ubuntu 18.04LTS Bionic: Its default gcc version is ok.
       * Ubuntu 16.04LTS Xenial: [Instructions](https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5) for installing gcc-7 in this version of Ubuntu.
+    * clang-4 or newer.
     * Windows: Visual Studio 2017 version 15.3 or newer.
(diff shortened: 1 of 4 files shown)


jolting:
Clang is still angry

jolting:
https://bugs.llvm.org/show_bug.cgi?id=33222

jolting:
We'll probably need to install clang-7.0 in the container from here:
http://apt.llvm.org

jlblancoc:
Ok, thanks for investigating the bug!
I'll try it in this branch...

jlblancoc:
clang-7 still fails with the same error... since it's a libstdc++ thing, obviously :-(

jlblancoc:
What do you think about a custom version of STD visit as a workaround?

jolting:
I was thinking the same thing.

jlblancoc:
@jolting I spent a few hours writing my own visit(), just to find out that it works fine on gcc but fails with the same error in clang (!!). 

I'm afraid the only solution at this point is disabling variant for clang for a __GLIBCXX__ older than the current version... :-(

jlblancoc:
@jolting feel free of merging if travis is happy after the last commit. Night!

jolting:
That's fair. Write variable is not as useful as read variant.

Good night.

PR #753, opened 2018-05-06
Use nested namespaces c++17

bergercookie:
Current PR transforms multiple namespace occurances to their C++17 nested namespaces counterpart.

* The python script used for this process can be found in scripts/find_replace_nested_namespaces.py
* Although 99% of times it gets it correct, there were cases where manual intervention was necessary. If we were to cover all of the cases and be extra confident about it we'd need to use a code aware tool and not a string-processing script
* The script couldn't operate on multiple files as it wasn't smart enough to identify the closing brackets. These files are included in the attached textfile.
* As nested-namespaces are a C++17 feature when compiling the project with gcc 7.2 I get loads of errors of this form:
```
/home/berger/mrpt/apps/mrpt-performance/perf-gridmaps.cpp:10:
/home/berger/mrpt/libs/maps/include/mrpt/maps/COccupancyGridMap2D.h:34:15: warning: nested namespace
      definition is a C++17 extension; define each namespace separately [-Wc++17-extensions]
namespace mrpt::maps
```
Adding the `-Wc++17-extensions` to cmake didn't help. Also I tried instructing CMake to use the 17 standard, but that's not supported before CMake 3.8.2. Obviously that's something we wouldn't want to get when compiling the project.
Any suggestion on how to solve this?

* Some of the files under `hwdrivers` were saved with dos line endings. These were converted to the standard Linux file ending by the script, thus the major changes reported by diff in these files.

[unmodified_files.txt](https://github.com/MRPT/mrpt/files/1977889/unmodified_files.txt)

P.S. @jlblancoc In case you're having second thoughts of merging this, don't hesitate to tell me. :-) 

---

I acknowledge to have:

(Notify: @MRPT/owners )


jolting:
-Wno-c++17-extensions

bergercookie:
>-Wno-c++17-extensions

Hmm, It actually works now, must have misplaced it the previous time I tried it. Thanks!

jolting:
-Wc++17-extensions enables it
-Wno-c++17-extensions disables it

PR #756, opened 2018-05-07
Add rdp for remote debugging

jlblancoc:
RDP debugging for AppVeyor builders.


jolting:
Any luck with this? I think the dependency walker would be useful thing to use.

jolting:
Looks like CWirelessPower.cpp causes hwdrivers to link again wlanapi.dll which for whatever reason isn't in the appveyor image.

jolting:
I suspect that windows server 2016 doesn't ship with wlan support.

PR #768, opened 2018-05-17
Disable SIMD by default if auto detect is off

jolting:


jlblancoc:
I was getting mad due to this: https://github.com/MRPT/mrpt/pull/768/commits/5db8a71801a1fbee020a026b43672e705fb43f79  (arm builds showed "Automatic" in the final cmake summary!)

Now all make sense. Merging.

Issue #770, opened 2018-05-18
Investigate wrong results for graph-slam optimizer (EDGE_SE2)


jlblancoc:
Problem: expected graph for the LevMarq optimizer in mrpt-graphslam doesn't match correct optimal (e.g. from G2O).

To fix:
* master: 
  * [x] SE(2)
  * [x] SE(3)
* mrpt-1.5
  * [x] SE(2)
  * [ ] SE(3) Won't fix. Use mrpt-2 if needed.

* [ ] Also: https://github.com/jlblancoc/tutorial-se3-manifold/issues/5

jlblancoc:
Probably the toughest-to-fix and subtlest bug in a while... I've been struggling for 1 entire week 😠 
A bug that has been around for, probably, 8 years (!!). 

I'll have to re-apply this fix to all kind of graphslam optimizers and to bundle adjustment. 
It'll be fixed in both, master and mrpt-1.5. 

@bergercookie : After this, all your graphslam code should work **amazingly** better. Testing against G2O now gives exactly the same results (optimal graphs up to machine precission).

bergercookie:
> Probably the toughest-to-fix and subtlest bug in a while... I've been struggling for 1 entire week angry
A bug that has been around for, probably, 8 years (!!).

Alright!!! Quite impressive, gonna test it out!

So where was the actual bug?

https://github.com/MRPT/mrpt/commit/8786bd6f76076ce667b1808bd92efab9718f67c1#diff-b2e9b4a71fbc6b6406855b6a2f1684f2L435 ?

jlblancoc:
That one, together with the use of H_map as a variable OUTSIDE the while(iter) loop... so all Hessian blocks were actually **accumulating** over iterations 🤕 

Will summarize the changes in a clean PR anyway.

jlblancoc:
Reopening after https://github.com/MRPT/mrpt/commit/609e607a8f27ad6969d854896b000f32ae944449. 

It turned out more robust formulation for graphslam Jacobians was needed to make things really robust.

To-do: https://github.com/jlblancoc/tutorial-se3-manifold/issues/5

jlblancoc:
Status of this issue: 
* SE(2) stuff is done (pending documentation in tech. report)
* SE(3) still to be done: `jacobian_dDinvP1invP2_depsilon()`, etc. 

WIP is in branch `test-alternative-slam2d-jacobs` for mrpt 2.0, mrpt-1.5 itself for 1.5.

PR #773, opened 2018-05-25
graph-slam improvements and fixes

jlblancoc:
## Changed apps/libraries

* mrpt-graphslam
* mrpt-graphs
* mrpt-io
* unit tests

## PR Description

* Solves #770  for `master`.
* CNetworkOfPoses: read/write format made compatible with G2O EDGE_SE2 
* CTextFileLinesParser can be bound to a generic std::istream
* Add an additional build stage to Travis to avoid timeouts.


jolting:
If I remember correctly, the docker image was only saved after stage 1.

CircleCI has built-in support for stages which is much nicer than travis, but it's $50/month for unlimited build minutes.

jlblancoc:
Correct! I'll go on with this PR without that last change, then open it as a new PR to avoid mixing things up.

PR #775, opened 2018-05-27
Travis 3 stages

jlblancoc:
Let's see if this works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,41 +20,49 @@ env:
 
 jobs:
   include:
-    - stage: Build Stage 1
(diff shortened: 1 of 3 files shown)


jolting:
The image name should have something like
full-clang-402c30972d64624872527226a88befbf447dab30-stage1
Stage 2 should pull $IMAGE-stage1 and save $IMAGE-stage2
Stage3 then starts from $IMAGE-stage2

I don't recommend using the image tag. It might make rebuilds tricky.

jolting:
Looks good now.

PR #783, opened 2018-06-05
WIP PR for JSON schema serialization [GsOC2018]

jlblancoc:
WIP: Addition of schema-based serialization methods


jlblancoc:
One more comment: @jolting , you are better aware of the timing for this project but, wouldn't be great to add unit testing right from the beginning?

jlblancoc:
>  to add unit testing

I just realized this would imply adding the json library as a CMake dependency, and to define the Json-specific SchemaArchive to mrpt-serialization. It would be ok to do so, as long as the dependency is optional and the whole thing keeps building without it.

jolting:
Rachit can create a mocking layer for all the unit tests that would implement the abstract base class and unit test that. This certainly wouldn't depend on Jsoncpp.

rachit173:
@jolting 
I am working on pimpl, and cleaning the merge. I am bit unfamiliar with unit testing, what is meant by mocking layer?

jolting:
You'll need implement the abstract base class with a mocking class that will facilitate inspection of the methods.
MRPT uses gtest
https://github.com/google/googletest/blob/master/googletest/docs/Primer.md

This might be slightly out of date
https://www.mrpt.org/Unit_testing_in_MRPT

Here's an example of the current serializeable unit test
https://github.com/MRPT/mrpt/blob/master/libs/serialization/src/CSerializable_unittest.cpp

rachit173:
@jlblancoc @jolting I have created a Mock class, but what kind of implemetation should be there, making an alias of jsoncpp would be difficult. Also I found [jsoncpp as simple header and cpp file](https://github.com/open-source-parsers/jsoncpp/wiki/Amalgamated). Can these be added to serialization library with the unittest?

rachit173:
Removed indirections using pimpl idiom and added unit test for json schema serialization The unit test is based on [jsoncpp as simple header and cpp file](https://github.com/open-source-parsers/jsoncpp/wiki/Amalgamated). I will rebase soon.

jolting:
Much better syntax.

rachit173:
What more changes are required for it to be ready for merge? And what ds are essential for serialization? I was currently thinking about most poses ds, some matrix ds.

jlblancoc:
Good progress!

However, @jolting and myself recommended you **not** including here for now any real JSON parser, but it's included (2.2k LOC) in  `libs/serialization/include/json/json.h` . Perhaps you didn't get the idea of what we meant with a "mocking  class"? 

@jolting: what do you think? should this be changed to a dummy mocking class just for testing?

rachit173:
Yes, actually confused about mocking class. Should I emulate the Json parser in the unit test mocking class, like by using map and strings? How does it check the functionality of schema serialization

jolting:
I think a dummy mocking class is necessary to write meaningful unit tests.

rachit173:
Is the dummy mocking class a class like CSchemeArchive, derived from CSchemeArchiveBase_impl? If that is the case , how should one mock the Json parser in the overriding functions?

jolting:
I would start with the simple things
```
out["foo"][1] = 1.0;
```
Just check in the mocking class that operator[](std::string_view) is called followed by operator[](size_t) and finally operator=(double) with the proper parameters. The mocking class is just part of the unit test and doesn't need to exist anywhere else in the code base.

rachit173:
Ok so it actually need not store 1.0,  test whether function calls don't produce errors. On it.

jolting:
Please cleanup some of the Git commits where you add then removed files. There is no sense in having dirty git commit history.

jlblancoc:
What's the state of this one?

libs/opengl/src/CCylinder.cpp
```diff
@@ -59,7 +59,42 @@ void CCylinder::render_dl() const
 
 #endif
 }
-
+void CCylinder::serializeTo(mrpt::serialization::CSchemeArchiveBase& out) const
+{
+	out["datatype"] = std::string(this->GetRuntimeClass()->className);
```
jlblancoc: What about defining a MACRO (define) to avoid repeating this long line in every class?

libs/opengl/src/CCylinder.cpp
```diff
@@ -59,7 +59,42 @@ void CCylinder::render_dl() const
...
+	out["slices"] = mSlices;
+	out["stacks"] = mStacks;
+	out["hasBottomBase"] = mHasBottomBase;
+	out["hasTopBase"] = mHasTopBase;
+}
+void CCylinder::serializeFrom(mrpt::serialization::CSchemeArchiveBase& in)
+{
+	uint8_t version = static_cast<int>(in["version"]);	// default is 0
+	if(static_cast<std::string>(in["datatype"]) ==
+		std::string(this->GetRuntimeClass()->className))	// match the class name
```
jlblancoc: idem here: I think it would be better to define a macro that contains code like this if(), but which, if the class name does not match, it raises an exception explaining the error. As it is now, the error would pass unnoticed (!), right?

jlblancoc:
@rachit173 : Please, see my last pair of comments and after that, we'll merge this one!

jlblancoc:
I have merged this PR. Then rebased and `push --force ` to remove commit of temporary files (CMakeCache.txt and Makefile in a sample directory!).

Now will do what I mentioned in the couple of comments above two days ago, since I had no feedback, and if travis is happy, we'll be able to merge to master.

PR #789, opened 2018-06-23
Add MAKE_ORGANIZED parameter

karnikram:
## PR Description
* Added MAKE_ORGANIZED parameter to T3DPointsProjectionParams to be able to generate an organized pcl point cloud from a 2D range image
* Appropriately set pcl cloud dimensions
* More of an "ad-hoc" solution, just to get pcl::IntegralNormalEstimation to work for my gsoc project. Please let me know if I need to make more changes.
---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/maps/include/mrpt/maps/CPointsMap.h
+++ b/libs/maps/include/mrpt/maps/CPointsMap.h
@@ -1200,6 +1200,10 @@ class PointCloudAdapter<mrpt::maps::CPointsMap>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
+	{
+	}
 	/** Get XYZ coordinates of i'th point */
 	template <typename T>
 	inline void getPointXYZ(const size_t idx, T& x, T& y, T& z) const
--- a/libs/maps/include/mrpt/maps/CSimplePointsMap.h
+++ b/libs/maps/include/mrpt/maps/CSimplePointsMap.h
@@ -177,6 +177,10 @@ class PointCloudAdapter<mrpt::maps::CSimplePointsMap>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
+	{
+	}
 	/** Get XYZ coordinates of i'th point */
 	template <typename T>
 	inline void getPointXYZ(const size_t idx, T& x, T& y, T& z) const
--- a/libs/obs/include/mrpt/obs/CObservation3DRangeScan.h
+++ b/libs/obs/include/mrpt/obs/CObservation3DRangeScan.h
@@ -44,15 +44,18 @@ struct T3DPointsProjectionParams
 	bool PROJ3D_USE_LUT;
 	/** (Default:true) If possible, use SSE2 optimized code. */
 	bool USE_SSE2;
-	/** (Default:true) set to false if you want to preserve the organization of
-	 * the point cloud */
+	/** (Default:true) set to false if your point cloud can contain undefined values */
 	bool MAKE_DENSE;
+	/** (Default:false) set to true if you want an organized point cloud */
+	bool MAKE_ORGANIZED;
+
 	T3DPointsProjectionParams()
 		: takeIntoAccountSensorPoseOnRobot(false),
 		  robotPoseInTheWorld(nullptr),
 		  PROJ3D_USE_LUT(true),
 		  USE_SSE2(true),
-		  MAKE_DENSE(true)
+		  MAKE_DENSE(true),
+		  MAKE_ORGANIZED(false)
 	{
 	}
 };
@@ -828,6 +831,11 @@ class PointCloudAdapter<mrpt::obs::CObservation3DRangeScan>
 		if (N) m_obj.hasPoints3D = true;
 		m_obj.resizePoints3DVectors(N);
 	}
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
+	{
+	}
+
 
 	/** Get XYZ coordinates of i'th point */
 	template <typename T>
--- a/libs/obs/include/mrpt/obs/CObservation3DRangeScan_project3D_impl.h
+++ b/libs/obs/include/mrpt/obs/CObservation3DRangeScan_project3D_impl.h
@@ -56,6 +56,9 @@ void project3DPointsFromDepthImageInto(
 	pca.resize(WH);  // Reserve memory for 3D points. It will be later resized
 	// again to the actual number of valid points
 
+	if(projectParams.MAKE_ORGANIZED)
+		pca.setDimensions(H,W);
+
 	if (src_obs.range_is_depth)
 	{
 		// range_is_depth = true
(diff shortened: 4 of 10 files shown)


libs/maps/include/mrpt/maps/CPointsMap.h
```diff
@@ -1200,6 +1200,10 @@ class PointCloudAdapter<mrpt::maps::CPointsMap>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
```
jlblancoc: I think this method does NOT need to be added here... it's enough to have the empty setDimensions() method in the PointCloudAdapter specialization for all MRPT classes, and only make it to have a real effect in the PCL specialization, right?

libs/maps/include/mrpt/maps/CSimplePointsMap.h
```diff
@@ -177,6 +177,10 @@ class PointCloudAdapter<mrpt::maps::CSimplePointsMap>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
```
jlblancoc: same here...

libs/obs/include/mrpt/obs/CObservation3DRangeScan.h
```diff
@@ -47,12 +47,16 @@ struct T3DPointsProjectionParams
...
+	bool MAKE_ORGANIZED;
+
 	T3DPointsProjectionParams()
 		: takeIntoAccountSensorPoseOnRobot(false),
 		  robotPoseInTheWorld(nullptr),
 		  PROJ3D_USE_LUT(true),
 		  USE_SSE2(true),
-		  MAKE_DENSE(true)
+		  MAKE_DENSE(true),
+		  MAKE_ORGANIZED(false)
```
jlblancoc: good!

libs/obs/include/mrpt/obs/CObservation3DRangeScan.h
```diff
@@ -828,6 +832,11 @@ class PointCloudAdapter<mrpt::obs::CObservation3DRangeScan>
 		if (N) m_obj.hasPoints3D = true;
 		m_obj.resizePoints3DVectors(N);
 	}
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
```
jlblancoc: same here...

libs/opengl/include/mrpt/opengl/CPointCloud.h
```diff
@@ -334,6 +334,10 @@ class PointCloudAdapter<mrpt::opengl::CPointCloud>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
```
jlblancoc: same here...

libs/opengl/include/mrpt/opengl/CPointCloudColoured.h
```diff
@@ -299,6 +299,10 @@ class PointCloudAdapter<mrpt::opengl::CPointCloudColoured>
 	inline size_t size() const { return m_obj.size(); }
 	/** Set number of points (to uninitialized values) */
 	inline void resize(const size_t N) { m_obj.resize(N); }
+	/** Does nothing as of now */
+	inline void setDimensions(const size_t& height, const size_t& width)
```
jlblancoc: same here...

jlblancoc: ah! Please, add this change to the docs/doxygen-pages/changelog.h list of changes...

libs/maps/include/mrpt/maps/CPointsMap.h
karnikram: I had to add the method to other specializations too because `project3DPointsFromDepthImageInto` gets instantiated using the non-pcl types too. For instance, [here](https://github.com/MRPT/mrpt/blob/970cf8b280b3b8e767292013aeeb01c4485dce5b/libs/obs/src/CObservation3DRangeScan.cpp#L1089).

libs/obs/include/mrpt/obs/CObservation3DRangeScan.h
karnikram: Forgot to add this,
MAKE_DENSE does not actually preserve the "organization" of the point cloud. It is only used to control whether or not the points can have nan / undefined values. If MAKE_DENSE is true, the cloud simply cannot contain undefined points and all the points contain finite values.

libs/opengl/include/mrpt/opengl/CPointCloudColoured.h
karnikram: Made the changes you requested!
And did a git commit --amend, to make it one commit. Hope that's okay.

libs/maps/include/mrpt/maps/CPointsMap.h
jlblancoc: > because project3DPointsFromDepthImageInto gets instantiated using the non-pcl types too.

Sure... but if I recall it right, it internally instantiates a PointCloudAdapter (PCA) of the required target type as template argument. Then, my point is: shouldn't it be `setDimensions()` a method of the PCA instead of the base, target types? In this way, only one PCA should be modified to add a dummy (empty)  `setDimensions()` method, the version for all MRPT classes. Only the specialization for PCL should have a `setDimensions()` method with real stuff inside.

Let me know if this makes now sense to you...

karnikram: Yes, that makes sense. I did try to do that at first, but I couldn't actually find the definition for the class template `PointCloudAdapter` adapter anywhere... The class declaration is [here](https://github.com/MRPT/mrpt/blob/deca6243a8ddd4b36ac450f1c4f9316a87f3fa0e/libs/opengl/include/mrpt/opengl/pointcloud_adapters.h#L38), but only specialization specific definitions have been written, I think.

Or maybe I'm wrong..

jlblancoc: Update: my fault. I didn't realize you added the methods to `PointCloudAdapter<mrpt::maps::CPointsMap>`, it seemed to me that the method was added to `CPointsMap` directly! So, it's good!

libs/obs/include/mrpt/obs/CObservation3DRangeScan_project3D_impl.h
```diff
@@ -11,6 +11,9 @@
 
 #include <mrpt/core/round.h>  // round()
 
+#include <pcl/point_cloud.h>
+#include <pcl/point_types.h>
+
```
jlblancoc: Ey! Red alarm here! :-)

We can't assume PCL is installed, since PCL is an **optional** dependency for MRPT.
I think you could remove those headers and all should build nice. Why did you add them? 
For the templates for build, it's enough to have forward declarations, and in any case, the PCL headers included **in the user code** (the caller that includes this file). 

So, please, remove those two includes.

jlblancoc:
Good. Let's wait for the travis result...

jlblancoc:
@jolting : Please, look at [this](https://travis-ci.org/MRPT/mrpt/jobs/396860356)... 

```
Error response from daemon: Get https://registry-1.docker.io/v2/: unauthorized: incorrect username or password
```

I think we didn't change anything in the docker area (?).

On the AppVeyor build, it seems it timed out (sigh)...

jolting:
@jlblancoc Travis isn't passing in env variables to PR. That's a setting on travis. I'm not so worried about leaking docker hub credentials in this way because that account is only used for builds. It's just a risk if we enable builds for PR from outside of the main repo.

PR #814, opened 2018-08-08
merge JSON-capable serialization methods

jlblancoc:
By @rachit173 for GSOC 2018. See #783 

Additional tasks to be done before merging: 
* [x] Helper macros
* [x] Add serialization for CPose* classes.
* [x] Add serialization for, at least, some CObservation* classes. Moved to #820 
* [x] Add real unit tests?
* [x] Add examples + update dox docs with the 2 ways to use JSON serialization ("raw" and "opaque").

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,6 +292,7 @@ include(cmakemodules/script_pcap.cmake REQUIRED)        # Check for the libpcap 
 include(cmakemodules/script_vtk.cmake REQUIRED)         # Check for VTK
 include(cmakemodules/script_clang_tidy.cmake REQUIRED)  # Clang tidy
 include(cmakemodules/script_cvd.cmake REQUIRED)         # Check for VTK
+include(cmakemodules/script_jsoncpp.cmake REQUIRED)     # Check for jsoncpp
 
 # ---------------------------------------------------------------------------
 #			OPTIONS
--- a/cmakemodules/DefineExamples.cmake
+++ b/cmakemodules/DefineExamples.cmake
@@ -70,6 +70,7 @@ IF(BUILD_EXAMPLES)
 	# === Depending on: serialization, io ===
 	SET(LIST_EXAMPLES_IN_THIS_DIR
 		# ----
+		serialization_json_example
 		serialization_stl
 		serialization_variant_example
 		# ----
--- a/cmakemodules/script_jsoncpp.cmake
+++ b/cmakemodules/script_jsoncpp.cmake
@@ -1,0 +1,15 @@
+
+SET(CMAKE_MRPT_HAS_JSONCPP 0)
+SET(CMAKE_MRPT_HAS_JSONCPP_SYSTEM 0)
+
+OPTION(DISABLE_JSONCPP "Forces NOT using JSONCPP, even if it could be found by CMake" "OFF")
+MARK_AS_ADVANCED(DISABLE_JSONCPP)
+
+IF(NOT DISABLE_JSONCPP)
+	find_package(jsoncpp QUIET)
+	if(jsoncpp_FOUND)
+		# cmake imported target: "jsoncpp_lib"
+		SET(CMAKE_MRPT_HAS_JSONCPP 1)
+		SET(CMAKE_MRPT_HAS_JSONCPP_SYSTEM 1)
+	endif()
+ENDIF()
--- a/cmakemodules/script_show_final_summary.cmake
+++ b/cmakemodules/script_show_final_summary.cmake
@@ -117,6 +117,7 @@ MESSAGE(STATUS " ______________________ OPTIONAL LIBRARIES _____________________
 SHOW_CONFIG_LINE_SYSTEM("Assimp (3D models)                  " CMAKE_MRPT_HAS_ASSIMP "[Version: ${ASSIMP_VERSION}]")
 SHOW_CONFIG_LINE_SYSTEM("ffmpeg libs (Video streaming)       " CMAKE_MRPT_HAS_FFMPEG "[avcodec ${LIBAVCODEC_VERSION}, avutil ${LIBAVUTIL_VERSION}, avformat ${LIBAVFORMAT_VERSION}]")
 SHOW_CONFIG_LINE_SYSTEM("gtest (Google unit testing library) " CMAKE_MRPT_HAS_GTEST )
+SHOW_CONFIG_LINE_SYSTEM("jsoncpp (JSON format serialization) " CMAKE_MRPT_HAS_JSONCPP "[Version: ${jsoncpp_VERSION}]")
 SHOW_CONFIG_LINE_SYSTEM("libjpeg (jpeg)                      " CMAKE_MRPT_HAS_JPEG)
 SHOW_CONFIG_LINE_SYSTEM("liblas (ASPRS LAS LiDAR format)     " CMAKE_MRPT_HAS_LIBLAS)
 SHOW_CONFIG_LINE       ("mexplus                             " CMAKE_MRPT_HAS_MATLAB)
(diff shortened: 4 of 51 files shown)


jlblancoc:
@rachit173 , @jolting : Right away after starting to complete the tasks above, I realize the implementation misses an important feature: the possibility of hierarchical access. 

For example, right now we can serialize in a "schema" a CPoint2D and a matrix. If we now would like to serialize a CPoint2DPDFGaussian (=mean CPoint2D + cov matrix), we could avoid repetition (which **should** be avoided!) by delegating the serialization of the two member variables to their corresponding current serialization code. 

However, the `CSchemeArchiveBase` does not allow for "namespaces", "hierarchy" or whatever you want to call it, that would prevent all the mix of all data members at the same level. What I would expect is something like (in JSON):
```
{ "mean": { "x" : 0, "y": 1, ... }, "cov" : { ... } }
```

I'm not sure that the current API:

```
CSchemeArchiveBase out(std::make_unique<CSchemeArchive<Json::Value>>(val));
```

allows for this without an important refactoring (?). 

I think one possible solution would be to add a new CSchemeArchiveBase `operator []` overriding the existing dict accessor (or a new method, I don't mind), which would return a `CSchemeArchiveBase` wrapper for a namespace, child of the one you are invoking it. 

I'm unfamiliar with the `jsoncpp` API so I'm first asking in case you see a fast way to do it before having to dig into it...

Cheers

jolting:
It should look something like this

```
out["mean"] = m_mean;
out["cov"] = m_cov;
```

Thats why this method exists:
```
    virtual CSchemeArchiveBase &operator=(mrpt::serialization::CSerializable&) = 0;
```

jlblancoc:
Brilliant! Well designed, I missed it. Will use it and add it to the docs

jlblancoc:
Hi @rachit173 , 

Once I did put hands on this, again I found doubts: 

1) These methods in mrpt-web, override two virtual methods:

https://github.com/MRPT/mrpt-web/blob/84988cab6d242f41c78f0e827ad89d556ed9283e/libs/web/include/mrpt/web/CSchemeArchive.h#L95-L104

but I cannot see the `asSerializableObject()` method in the base class `mrpt::serialization::CSchemeArchiveBase_impl` (!!). See the latest version of the header in MRPT [here](https://github.com/MRPT/mrpt/blob/dev-json-schemas/libs/serialization/include/mrpt/serialization/CSchemeArchiveBase.h).

2) Then, apparently only one of these two methods are invoked:

```
protected:
 void ReadObject(CSchemeArchiveBase& out, CSerializable& obj);
 void WriteObject(CSchemeArchiveBase& out, CSerializable& obj);
```
??

3) The method @jolting mentioned above `virtual CSchemeArchiveBase &operator=(mrpt::serialization::CSerializable&) = 0;` is, if I got it right, designed to take a CSerializable object and convert it to a schema. 
However, here: https://github.com/MRPT/mrpt-web/blob/84988cab6d242f41c78f0e827ad89d556ed9283e/libs/web/include/mrpt/web/CSchemeArchive.h#L97

It calls the `ReadObject()` method... I would prefer calling to that "write" an object to a schema stream, rather than the other way around. Don't you agree?

4) In connection with this doubt: if one of the read/write methods, the CSerializable object should be passed as a `const` reference, to explicitly stating which one means "write" and which one "to read".

5) And finally: how is an object read from a CSchemaArchive, if the `ReadObject()` is protected and, as far as I see, not exposed at all? What I need is the opposite operation to `out["mean"] = mean;`, that is, something like: `out["mean"].ReadObject(mean);` or alike (?). 

Did you missed to pull request some new commits after my last merge? I couldn't see any new commit on your published MRPT fork...

cc: @jolting , anything I missed??

jolting:
1. I think it's here. https://github.com/MRPT/mrpt/blob/5660688ec875de8c0a2923443fbba1f9d60c7bfa/libs/serialization/include/mrpt/serialization/CSchemeArchiveBase.h#L46

4. Yes agreed. It needs to be a const reference.

ReadObject and WriteObject are unnecessary aliases for 
obj.serializeTo(out) and obj.serializeFrom(in)
I see the indirection unnecessary and confusing.

I think something like this could work:
m_mean = archive["mean"].as<CPose2D>().

rachit173:
Ignore my current MRPT fork it has some serious errors since I lost my local clone on my computer( my hard disk got corrupted a few days ago :-( , lost some unpushed changes). I will create a new fork from current MRPT repo and look into your suggestions.

jolting:
> I think it would be a good moment to add optional support for libjsoncpp 

I think we just need to move CSchemeArchive.h from mrpt-web to mrpt-serialization.

The interesting part is that it's totally templated, so it's already optional.

`CSchemeArchive<Json::Value>`

jlblancoc:
I think I'm still missing something on how to use the new scheme archive... please, take a look at my last commit (which fails to build) here: 

https://github.com/MRPT/mrpt/commit/f335324b41e33a16465261e74fcde10aa85e1632#diff-93c34139d8003f29db67d596cbd8f980R162

That aside: we now have jsoncpp as an optional dep as a cmake imported target and all is right. 

For now only used in the test, but when we manage it to work there, my idea is to expose new functions to read/write mrpt objects to json. Normal functions, with the body in a cpp so the jsoncpp dep is hidden to the user.

jlblancoc:
Funny fact: the code builds and runs perfectly under MSVC2017, with jsoncpp 1.8 and 1.7.2... 
So it's only something that affects gcc. Will investigate it...

jlblancoc:
@jolting : please, take a look at my last commits. I successfully created some real JSON-scheme unit tests for serialization. I don't have time to continue debugging what's wrong with de-serialization. 

```
make run_tests_mrpt_serialization
[  0%] Built target DocumentationFiles
[  9%] Built target mrptgtest
[ 18%] Built target mrpt-core
[ 18%] Built target mrpt-random
[ 27%] Built target mrpt-rtti
[ 27%] Built target mrpt-containers
[ 27%] Built target mrpt-expr
[ 45%] Built target mrpt-system
[ 45%] Built target mrpt-serialization
[ 45%] Built target mrpt-config
[ 54%] Built target mrpt-io
[ 72%] Built target mrpt-math
[ 72%] Built target mrpt-bayes
[100%] Built target mrpt-poses
[100%] Built target test_mrpt_serialization
[==========] Running 8 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 6 tests from Serialization
[ RUN      ] Serialization.STL_stdvector
[       OK ] Serialization.STL_stdvector (0 ms)
[ RUN      ] Serialization.STL_stdmap
[       OK ] Serialization.STL_stdmap (0 ms)
[ RUN      ] Serialization.STL_complex_error_type
[       OK ] Serialization.STL_complex_error_type (0 ms)
[ RUN      ] Serialization.vector_custom_type
[       OK ] Serialization.vector_custom_type (0 ms)
[ RUN      ] Serialization.vector_shared_ptr
[       OK ] Serialization.vector_shared_ptr (0 ms)
[ RUN      ] Serialization.CustomClassSerialize
[       OK ] Serialization.CustomClassSerialize (0 ms)
[----------] 6 tests from Serialization (0 ms total)

[----------] 2 tests from SchemaSerialization
[ RUN      ] SchemaSerialization.JSON_archive
[       OK ] SchemaSerialization.JSON_archive (0 ms)
[ RUN      ] SchemaSerialization.JSON_raw
unknown file: Failure
C++ exception with description "in Json::Value::resolveReference(key, end): requires objectValue" thrown in the test body.
[  FAILED  ] SchemaSerialization.JSON_raw (1 ms)
```

Probably I simply don't use the proper way to initialize a Json::Value from a string representation, [here](https://github.com/MRPT/mrpt/commit/5bc08e25f29cefef3b4f14b22d84f3e751596575#diff-93c34139d8003f29db67d596cbd8f980R52) (line 52).

Any hint would be more than welcome!

BTW: I solved the uint64 issue in a way that is not clearly portable to other "schemes" (apart of jsoncpp), but I realized the code is **already** not portable, since the generic template already invoked jsoncpp-specific method such that `asUInt64()`, etc. Not a big deal for me, though...

jlblancoc:
Nevermind, I found it! I'll push more changes later. Cheers

jlblancoc:
In theory, all is done in this PR. If travis is green after the last commit, we can merge it!

PR #816, opened 2018-08-12
TTimeStamp and mrpt::system::Clock

jlblancoc:
This PR will cover #423. 

Tasks and discussion: 
* [x] Convert TTimeStamp to a class with setter/getter members. See #423 . Nope: It seems better to keep the plain functions API.
* [x] Provide C++11 std::chrono-like clock interface. 
* [x] Add more unit tests.

The std::chrono part is, I think, mostly done in mrpt::system::Clock. However, I wouldn't like to remove the current `mrpt::system::now()` since it's well tested on all OSes to provide a high-resolution timer, which is not ensured by the C++11 std::chrono API...
We could just change `std::chrono::system_clock` [here](https://github.com/MRPT/mrpt/blob/3e53bdf4bbf8da532735bc2510e6599c5b2bdc45/libs/system/include/mrpt/system/Clock.h#L29) with `high_resolution_clock`, but we would be in the hands of the C++ library implementation (!), so I wouldn't vote for that.

I even thought of defining `using TTimeStamp = mrpt::system::Clock::time_point` (to keep a certain level of backwards compatibility), then move all functions inside `TTimeStamp` like:

```
class Clock
{
   public:
// was: using time_point = std::chrono::time_point<Clock>;
   struct time_point : public std::chrono::time_point<Clock>
   {
      double toDouble() const;
      double timeDifferenceTo(const std::chrono::time_point<Clock> &o);
      static std::chrono::time_point<Clock> fromTime_t(time_t value); 
      //...
   };
``` 

Thoughts on this?


jlblancoc:
I just re-read @jolting 's comment from #423 : 

> I wonder if it would just be better to refactor code that uses TTimeStamp to directly use std::chrono::time_duration.

which is along the line of making TTimeStamp equivalent to the new clock's time_duration. 
The only caveat I have is, as said above, the high-resolution timer. Perhaps we could move the `Clock::now()` implementation to the cpp file and use the current, ugly but tested implementation?

jolting:
Moving clock to core was necessary for the serialization library. I'm not necessarily happy with that change. Mrpt serialization probably should treat time like other serializable objects. Of course deserialization needs to be backwards compatible.

jlblancoc:
I see. It's fine. 
Only, the very few entities in core directly are defined inside mrpt namespace, not mrpt::core... again, for a better backward compatibility.

Clock should be moved to mrpt as well, I guess, but that's of no importance now!

I'm on the phone now. Just pushed wip work trying to move the old now()...

libs/core/src/Clock.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+	// Convert to TTimeStamp 100-nanoseconds representation:
+	return uint64_t(tim.tv_sec) * UINT64_C(10000000) +
+		   UINT64_C(116444736) * UINT64_C(1000000000) + tim.tv_nsec / 100;
+}
+
+Clock::time_point Clock::now() noexcept
+{
+	// TODO: Fix this one!
+	return time_point(getCurrentTime());
```
jolting: You need to do this explicitly.
```
return time_point(duration(getCurrentTime()));
```

jlblancoc:
I'm porting everything such that TTimeStamp is simply a Clock::time_point. I'll push in a while... please review it to check if you see something I might have missed.

I'm more happy with the current state of the date/time API: there were many opportunities for mistakes with the original method of using the same type (TTimeStamp) for: UTC time, local time (!), and durations (!!).

PS: I `push --force` the branch to clean up the commit tree, I guess you might find trouble updating, sorry...

jlblancoc:
Let's wait for Travis...

Btw: I left INVALID_TIMESTAMP for backward compatibility but I still agree in using optimal wherever it makes sense.

apps/RawLogViewer/main_imports_exports.cpp
```diff
@@ -1688,8 +1689,8 @@ void xRawLogViewerFrame::OnMenuItemImportBremenDLRLog(wxCommandEvent& event)
 	std::string line;
 
 	mrpt::system::TTimeStamp cur_timestamp = mrpt::system::now();
-	const mrpt::system::TTimeStamp time_steps =
-		mrpt::system::secondsToTimestamp(0.25);  // why not? ;-)
+	const auto time_steps =
+		std::chrono::microseconds(static_cast<uint64_t>(0.25 * 1e6));
```
jolting: This might be a bit convoluted. std::chrono::milliseconds(250) is much clearer.

libs/hwdrivers/src/CIMUIntersense.cpp
```diff
@@ -190,11 +180,8 @@ void CIMUIntersense::doProcess()
 		else
 			AtUI = nowUI - m_timeStartUI;
 
-		// mrpt::system::TTimeStamp AtDO = mrpt::system::secondsToTimestamp(
-		// AtUI * 1e-6 /* Board time is usec */ );
-		mrpt::system::TTimeStamp AtDO =
-			mrpt::system::secondsToTimestamp(AtUI /* Board time is sec */);
-		obs->timestamp = m_timeStartTT + AtDO;
+		/* Board time is sec */
+		obs->timestamp = m_timeStartTT + std::chrono::milliseconds(AtUI * 1e-3);
```
jolting: std::chrono::microseconds(AtUI)

libs/hwdrivers/src/CKinect.cpp
```diff
@@ -546,8 +547,8 @@ void CKinect::getNextObservation(
 #if MRPT_HAS_KINECT_FREENECT
 
 	static const double max_wait_seconds = 1. / 25.;
-	static const TTimeStamp max_wait =
-		mrpt::system::secondsToTimestamp(max_wait_seconds);
+	static const auto max_wait = std::chrono::milliseconds(
+		static_cast<uint64_t>(max_wait_seconds * 1e3));
```
jolting: std::chrono::seconds

jlblancoc: Doesn't it accept integers only?

apps/RawLogViewer/main_imports_exports.cpp
jlblancoc: This one is crystal clear, yep!

jlblancoc:
Closed with a manual rebase + merge with:
* 108cec57377b3204bb866d6ea98d613abc39f29e
* c1e9f5685e4cb55a0a6b90270e4ab5ff316fb76e
* 6202c4131614222801e8fc7e0fc163d3c4262c61

Issue #810, opened 2018-07-27
Problem with 3DObservations (mrpt1.5 + Windows only?)


jlblancoc:
It seems 3D range observations, when using mrpt1.5 and MSVC, gives wrong point cloud coordinates. 

Added a test dataset here: 3f03c94195d4f90dbefbe2c72979b4d704038849

jlblancoc:
to-do: Cherry-pick part of 33e56689811b7092b7e749ea08746568a4d9e1ba into mrpt-1.5 to continue debugging this issue.

jlblancoc:
Interesting: the new test fails also for GCC, when built in i386 w/o SSE2 (as detected in the PPA build server):

```
[ RUN      ] CObservation3DRangeScan.LoadAndCheckFloorPoints
/<<PKGBUILDDIR>>/libs/obs/src/CObservation3DRangeScan_unittest.cpp:276: Failure
Expected: (bin_count[14]) >= (12000), actual: 0 vs 12000
/<<PKGBUILDDIR>>/libs/obs/src/CObservation3DRangeScan_unittest.cpp:277: Failure
Expected: (bin_count[18]) <= (700), actual: 1866 vs 700
/<<PKGBUILDDIR>>/libs/obs/src/CObservation3DRangeScan_unittest.cpp:278: Failure
Expected: (bin_count[19]) <= (20), actual: 3101 vs 20
```

PR #821, opened 2018-08-25
Add unit test for 3D range scans

jlblancoc:
Add new unit test with sample dataset to check that a 3D range scan of the floor actually has the correct coordinates after unprojecting depth -> 3D. 

Instrumental to help debugging #810 .


jlblancoc:
Interestingly, this new unit tests revealed a hidden bug (?): when building in the minimal configuration without zlib, we cannot read gz-compressed files, despite there's an embedded version of gzlib which should do the work...

PR #824, opened 2018-08-29
Mrpt graphslam improvements

bergercookie:
## Changed apps/libraries

* mrpt_graphslam (mostly)

## PR Description

PR includes a parital cleanup of the mrpt_graphslam library. This mostly includes the newer C++11 syntax and constructs (see https://github.com/MRPT/mrpt/issues/697)

---

I acknowledge to have:

(Notify: @MRPT/owners )

### DIFF
--- a/libs/graphslam/include/mrpt/graphslam/CGraphSlamEngine_impl.h
+++ b/libs/graphslam/include/mrpt/graphslam/CGraphSlamEngine_impl.h
@@ -7,8 +7,7 @@
    | Released under BSD License. See details in http://www.mrpt.org/License |
    +------------------------------------------------------------------------+ */
 
-#ifndef CGRAPHSLAMENGINE_IMPL_H
-#define CGRAPHSLAMENGINE_IMPL_H
+#pragma once
 
 #include <mrpt/system/filesystem.h>
 #include <mrpt/obs/CObservationOdometry.h>
@@ -2801,4 +2800,3 @@ void CGraphSlamEngine<GRAPH_T>::getDeformationEnergyVector(
 }
 }  // namespace mrpt::graphslam
 
-#endif /* end of include guard: CGRAPHSLAMENGINE_IMPL_H */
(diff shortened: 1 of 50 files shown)


jlblancoc:
Awesome, thanks for not forgetting! ;-)

Just curious: why do you prefer explicitly defining dtors as `=default` instead of simply not declaring them? I **think** all base classes and data members are "trivial" (not raw pointers or complexities alike), right? Probably you did it for some reason...

It seems we (again) have the Docker auth problem within Travis for **this** PR... Will test the code manually (tomorrow?) and merge then.

Logrus:
Probably just for readability?
Clang-tidy will try to do this:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-equals-default.html
and 
https://clang.llvm.org/extra/clang-tidy/checks/readability-deleted-default.html

jlblancoc:
Thanks @Logrus !

Indeed, that justifies using default instead of an empty body. 

But then my next question is: wouldn't it be more compact and totally equivalent just NOT defining any ctor/dtor at all if all data members are trivial? It would make the code a little shorter... ;-)

libs/graphslam/include/mrpt/graphslam/CGraphSlamEngine_impl.h
```diff
@@ -2575,16 +2532,14 @@ template <class GRAPH_T>
...
-	using namespace mrpt::gui;
 
 	MRPT_START;
 	MRPT_LOG_DEBUG_STREAM("In initializeSlamMetricVisualization...");
 	ASSERTDEB_(m_visualize_SLAM_metric);
 
 	// initialize the m_win_plot on the stack
-	m_win_plot = new CDisplayWindowPlots(
-		"Evolution of SLAM metric - Deformation Energy (1:1000)", 400, 300);
+	m_win_plot.reset(new mrpt::gui::CDisplayWindowPlots(
```
jolting: I'd prefer `std::make_shared<mrpt::gui::CDisplayWindowPlots>("Evolution of SLAM metric - Deformation Energy (1:1000)", 400, 300))`;

jolting: I of course meant std::make_unique

bergercookie:
> Just curious: why do you prefer explicitly defining dtors as =default instead of simply not declaring them?

I slightly prefer `default` just because it makes it explicit

PR #835, opened 2018-09-30
Various ExprTk updates.

ArashPartow:
Fixes to the globbing
Minor updates and cleanups

**Note:** I couldn't find the devel branch - has it been removed?

### DIFF
--- a/libs/expr/include/mrpt/otherlibs/exprtk.hpp
+++ b/libs/expr/include/mrpt/otherlibs/exprtk.hpp
@@ -619,7 +619,14 @@ namespace exprtk
                return false;
          }
 
-         return (d_itr == data_end) && (p_itr == pattern_end);
+         if (d_itr != data_end)
+            return false;
+         else if (p_itr == pattern_end)
+            return true;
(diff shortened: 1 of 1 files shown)


jlblancoc:
Thanks! 

> Note: I couldn't find the devel branch - has it been removed?

Yes, it's true... I actually don't have a clear plan on how it could be possible to keep `devel` active on the long term, without having to be constantly syncing back and forth with `master`... CI checks on PRs against `master` are good enough, I guess. 

BTW: Just out of curiosity, what does your **long** version number suppose to mean? :-) Perhaps a camouflaged timestamp?

ArashPartow:
@jlblancoc  the versioning of the library is similar to that used by Knuth for TeX. 

In this case the number is [e](https://en.wikipedia.org/wiki/E_(mathematical_constant)), with the idea being that the library will gradually reach perfection asymptotically.  :D

jolting:
Neat, one digit at a time. 
Might be fun to do one with a taylor series.
1+1+1/2 + 1/6 +1/24 + 1/120+ 1/720 + 1/5040 + 1/40320+1/362880+1/3628800
Perhaps your version 2.0 could be e^2.

PR #838, opened 2018-10-06
OpenCV compatible pixel depth in CImage

jolting:


jlblancoc:
Awesome! Will check it later with more time, but it looks great.... it's something that needed to be done sometime!

jolting:
Makes ros comaptibility easier. You wouldn't believe how many intensity images are 16-bit.

jlblancoc:
> You wouldn't believe how many intensity images are 16-bit....

... He Couldn't Believe It! Click Here To Find It Out!! ;-)

Cool. The only caveat is that uint8_t is assumed in a number of places. I'll add a couple of asserts just to be safe before merging.

jlblancoc:
Hmm... truncation or [Mu-Law](https://en.wikipedia.org/wiki/%CE%9C-law_algorithm) seems good options that may both be useful depending on the case...

I just pushed a couple of commits. If all Circle-CI builds pass, we could merge.

jolting:
Did this compile on Windows? the enum shouldn't explicitly convert to int32_t.

PR #842, opened 2018-10-13
WIP: Add clang linter to circle-ci

jlblancoc:


jolting:
Does this work?

jolting:
Is that output right?

jlblancoc:
> Is that output right?

it should be now! https://github.com/MRPT/mrpt/pull/842/commits/11646b1eaf04c962a66b6a3f149ca4ce711dd587

PR #843, opened 2018-10-14
Fix 32bit overflow error

jlblancoc:
Fix errors like this (in Travis CI):

/mrpt/libs/img/src/CImage.cpp:101:8: error: case value evaluates to 2147483680, which cannot be narrowed to type 'int32_t' (aka 'int') [-Wc++11-narrowing]
                case IPL_DEPTH_32S:


jlblancoc:
Apparently related to: https://github.com/opencv/opencv/issues/8192

Issue #800, opened 2018-07-16
Export build flags via MRPTConfig.cmake


jlblancoc:
See discussion: 
https://github.com/mrpt-ros-pkg/mrpt_navigation/issues/102

jlblancoc:
This one would actually be automatically solved via #602, so it can be closed.

PR #844, opened 2018-10-14
WIP: MSVC appveyor fixes

jlblancoc:
Ongoing PR with MSVC fixes in AppVeyor (not detected in local builds!)

### DIFF
--- a/libs/core/CMakeLists.txt
+++ b/libs/core/CMakeLists.txt
@@ -8,5 +8,8 @@ define_mrpt_lib(
 	)
 
 if(BUILD_mrpt-core)
-
+	# Fix a MSVC binary-breaking compatibility in MSVC 2017 15.8:
+	if (MSVC)
+		target_compile_definitions(mrpt-core PUBLIC -D_ENABLE_EXTENDED_ALIGNED_STORAGE)
+	endif()
 endif()
--- a/libs/core/include/mrpt/core/aligned_allocator.h
+++ b/libs/core/include/mrpt/core/aligned_allocator.h
@@ -21,11 +21,6 @@
 #define MRPT_MAX_ALIGN_BYTES 16
 #endif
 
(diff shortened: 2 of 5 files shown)


Wehrdo:
When building on MSVC 15.8.7, [this define](https://github.com/MRPT/mrpt/blob/90d18e498742335d9ccf4691ac4a17d04fecaa6b/libs/core/include/mrpt/core/aligned_allocator.h#L26) is not doing it for me. I had to add it as a project-level define in [CMakeLists.txt](https://github.com/MRPT/mrpt/blob/msvc-appveyor-fixes/CMakeLists.txt#L373).

There's also an issue with [this typename as function default argument](https://github.com/MRPT/mrpt/blob/msvc-appveyor-fixes/libs/graphslam/include/mrpt/graphslam/levmarq.h#L83), which is apparently [a known bug](https://developercommunity.visualstudio.com/content/problem/311059/vc-158-typename-in-templatized-function-default-va.html) in 15.8 that requires a workaround.

With those two fixes, I can build successfully on 15.8.

jlblancoc:
Thanks! Could you please open a PR against this branch? 

I wanted to avoid defining more global symbols at the CMake level, since then they must be also duplicated in the MRPTConfig.cmake.in file, but if there's no other way to go... 

At least until we address #800 ...

jlblancoc:
Unbelievable... MS released a new version of MSVC (15.8.8) today... but the regression is still there!

Just fixed those two errors on my own way with: 385742d8bbbc7bd88c289bb085cb750006b81d2d

If AppVeyor is happy, this PR can be merged to master, ok? (cc: @jolting ... your turn to merge if it's fine... bedtime in Europe ;-)

jlblancoc:
I'll merge this one since the patch fixes some MSVC errors in AppVeyor. There are new unit test failures, but we can investigate them in a new PR...

Issue #852, opened 2018-10-28
CImageGrabber_OpenCV does not work with OpenCV 3.4.3 and Windows 10


Wehrdo:
Attempting to open a webcam using [CCameraSensor](http://mrpt.ual.es/reference/devel/classmrpt_1_1hwdrivers_1_1_c_camera_sensor.html) with the suggested code sample will connect to the camera, but fail to retrieve frames.

```cpp
CCameraSensor cam;
const string cam_config_str =
    "[CONFIG]\n"
    "grabber_type=opencv\n";
CConfigFileMemory cam_config(cam_config_str);
cam.loadConfig(cam_config, "CONFIG");
cam.initialize();
auto obs = cam.getNextFrame();
```

It looks like two things have conspired to cause failure.

1. CImageGrabber_OpenCV uses the "legacy" C API for video capture, which no longer supports DirectShow --- instantiating a camera with the above code prints this out in the console:
`[ WARN:0] cvCreateFileCaptureWithPreference: backend DSHOW doesn't support legacy API anymore.`
(OpenCV cycles through backends and attempts them [[source]](https://github.com/opencv/opencv/blob/e6dd9a78aee8ccd217114f2b530e9772cdb6eb81/modules/videoio/src/videoio_c.cpp#L37))

2. Not sure what the root cause is, but the Video for Windows (VFW) backend no longer works with OpenCV on recent versions of Windows 10. I'm apparently not the only one with [this issue](https://github.com/opencv/opencv/issues/11810).

It looks like the solution would be updating CImageGrabber_OpenCV to use the `cv::VideoCapture` C++ API, since it can use DSHOW on Windows.

To verify these conclusions, connecting to my camera with `cv::VideoCapture(0)` on OpenCV 3.4.3 works fine, but with `cv::VideoCapture(0 + CAP_VFW)` results in the same behavior as with MRPT: Camera successfully opens, but frames fail to retrieve. Also, stepping through the MRPT code, I can see that the default `CAMERA_CV_AUTODETECT` resolves to using the VFW backend.

jlblancoc:
Thanks for nailing it down! We'll fix it...

jlblancoc:
@Wehrdo let me know if the code in the branch [opencv-fixes](https://github.com/MRPT/mrpt/tree/opencv-fixes) does work now... 

For some reason I upgraded to opencv 3.4.3 and now I have some problems due to (apparently unrelated) missing DLLS (MF.DLL, etc.) 😞

[...]

PR #853, opened 2018-10-29
WIP: Opencv & mrpt-vision fixes

jlblancoc:
Connected to: #852 & #120


Wehrdo:
@jlblancoc This branch is able to retrieve camera frames for me now, but they seem to get corrupted at [this line](https://github.com/MRPT/mrpt/blob/opencv-fixes/libs/hwdrivers/src/CCameraSensor.cpp#L1102). Up until that point, the data in `obs->image.img` looks correct (width, height, etc.), but after the call to `setSensorPose`, the data gets clobbered. When using the ffmpeg grabber and loading a file, this doesn't happen, even though `setSensorPose` is still called.

Could the use of `cv::Mat` for `capImg` be causing the memory for the frame to be released upon the destruction of `capImg`? I'm not too familiar with how IplImage handles memory.

jlblancoc:
It's fixed now!

PR #859, opened 2018-11-10
Use modern CMake exported targets

### LINKED ISSUE #602: Use CMake export
Think whether it's worth changing the current `MRPTConfig.cmake` to use CMake standard "EXPORT" command. 
* Pros: should make it simpler to work with static vs dynamic libs (without all the ad-hoc CMake code currently present).
* Cons: well... the current solution "just works", despite being less elegant...

Inspirational example [here](https://github.com/RainerKuemmerle/g2o/compare/Export_cmake_Configuration#diff-dfdf08900eb0292ea317b9f75ef332ed).

jlblancoc:
This closes #602. 
Note that the update breaks compatibility with existing code finding MRPT (pre 2.0) like:

```
find_package(MRPT REQUIRED slam vision [...])
...
target_link_libraries(mytarget  ${MRPT_LIBS})
```

which should now be updated to:

```
# Find MRPT libraries:
find_package(mrpt-poses)
find_package(mrpt-gui)

# Define your own targets:
add_executable(myapp  main.cpp)

# Link against MRPT: this will also add all required flags,
# include directories, etc.
# Note: the namespace prefix `mrpt::` is optional.
target_link_libraries(myapp
  mrpt::mrpt-poses
  mrpt::mrpt-gui
)
```

Also added a doxygen page describing all this: [here](https://docs.mrpt.org/reference/cmake-config-targets/mrpt_from_cmake.html).

It was about time for this update: the modern approach using CMake targets allows all build flags to automatically propagate to the user via the PUBLIC/INTERFACE interface definitions. I'm sure it will be a win in the long term. 

Comments?


jlblancoc:
To do: Fix Find Eigen3 in the case of minimal deps (eigen as external project)

PR #872, opened 2018-12-03
allow loadFromBitmap() origin to be outside of the image

jlblancoc:
## Changed apps/libraries

* mrpt-maps

### DIFF
--- a/libs/maps/include/mrpt/maps/COccupancyGridMap2D.h
+++ b/libs/maps/include/mrpt/maps/COccupancyGridMap2D.h
@@ -1066,25 +1066,26 @@ class COccupancyGridMap2D : public CMetricMap,
 	 * \param file The file to be loaded.
 	 * \param resolution The size of a pixel (cell), in meters. Recall cells are
 	 * always squared, so just a dimension is needed.
-	 * \param xCentralPixel The `x` coordinate (0=first, increases <b>left to
-	 * right</b> on the image) for the pixel which will be taken at coordinates
-	 * origin (0,0). (Default: the center of the image)
-	 * \param yCentralPixel The `y` coordinate (0=first, increases <b>BOTTOM
-	 * upwards</b> on the image) for the pixel which will be taken at
-	 * coordinates origin (0,0). (Default: the center of the image)
-	 * \return False on any error.
-	 * \sa loadFromBitmap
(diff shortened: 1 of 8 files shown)


libs/maps/include/mrpt/maps/COccupancyGridMap2D.h
```diff
@@ -1066,25 +1066,24 @@ class COccupancyGridMap2D : public CMetricMap,
...
+	 * coordinates for the pixel which will be taken at the origin of map
+	 * coordinates (0,0). (Default: the center of the image) \return False on
+	 * any error. \sa loadFromBitmap
 	 */
 	bool loadFromBitmapFile(
-		const std::string& file, float resolution, float xCentralPixel = -1,
-		float yCentralPixel = -1);
+		const std::string& file, float resolution,
+		std::optional<mrpt::math::TPoint2D> origin =
+			std::optional<mrpt::math::TPoint2D>());
```
jlblancoc: I haven't found a less verbose way to express this with `std::optional`... any better idea? (cc: @jolting )

jolting: Add two overloaded functions and make the std::optional implementation private.
```
bool COccupancyGridMap2D::loadFromBitmap(
	const mrpt::img::CImage& imgFl, float res,
	mrpt::math::TPoint2D origin)
{
 return loadFromBitmap_impl(imgFl, res, origin);
}

bool COccupancyGridMap2D::loadFromBitmap(
	const mrpt::img::CImage& imgFl, float res)
{
 return loadFromBitmap_impl(imgFl, res);
}
```

Then do this

```
const bool grid_loaded_ok = m_Ocgridmap.loadFromBitmapFile(
  this->m_insertOptions_common->GMRF_gridmap_image_file,
  this->m_insertOptions_common->GMRF_gridmap_image_res,
  {
    this->m_insertOptions_common->GMRF_gridmap_image_cx,
    this->m_insertOptions_common->GMRF_gridmap_image_cy
  });
```

jlblancoc: Thanks! 
But then... using `optional` here doesn't make sense if it does not make things easier to the caller from the public API, sigh... I probably had higher expectations ;-)
Still, optional may be useful for structs of many input/output parameters and alike, but not for optional function arguments, I guess...

PR #877, opened 2018-12-06
Add CObservationPointCloud

jlblancoc:
## Changed apps/libraries

* mrpt-maps

Add a new CObservation class, comprising a generic pointcloud "map".


jolting:
An explicit converting constructor from CObservationRangeScan3D to CObservationPointCloud would be nice.

The implicit CPointMap generation embedded in CObservationRangeScan3D was always a bit burdensome IMHO. I'd much rather convert explicitly when I want a CPointMap.

PR #911, opened 2019-04-10
ROS2 no longer uses catkin for the build system

jolting:
Disable octomap until ROS2 supports it

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -29,13 +29,12 @@
   <depend>wxwidgets</depend>
   <depend>zlib</depend>
   <depend>assimp-dev</depend>
-  <depend>octomap</depend>
+  <!-- <depend>octomap</depend> -->
 
   <doc_depend>doxygen</doc_depend>
 
   <!-- Minimum entries to release non-catkin pkgs: -->
   <buildtool_depend>cmake</buildtool_depend>
-  <exec_depend>catkin</exec_depend>
   <export>
     <build_type>cmake</build_type>
     <rosdoc config="doc/rosdoc.yaml" />


jolting:
ROS2 should break a bunch of other things, so I'm thinking mrpt ROS2 modules use mrpt2 as well.

ROS2 bumps up support to C++14 and c++17 is being considered. I believe MRPT c++17 compiler flags should work on all supported ROS2 platforms.

ROS2 hasn't quite hit the same release cadence yet. I'm not sure if an LTS release is coming soon. The latest release goes EOL in December, so long term support is unnecessary. It might be that by the time ROS2 is stabilized enough for an LTS then MRPT1 might be deprecated.

jlblancoc:
Hi, 

> ROS2 bumps up support to C++14 and c++17 is being considered. 

Hmm... ROS1 Melodic [already supports](https://github.com/MRPT/mrpt/wiki/MRPT-ROS-packages) C++14, so *I hope* the next one will allow c++17!

> I believe MRPT c++17 compiler flags should work on all supported ROS2 platforms.

Hope so! In fact, it already works with ROS1 Melodic (MRPT2 (c++17) built against ROS1, I mean). No ROS1 API seems to break due to the C++17 flag :+1:

For reference, today I saw this:
https://github.com/ros-drivers/velodyne/issues/165#issuecomment-482228956

that's a good option: to keep one single git repo for ROS1/ROS2 MRPT packages. 


On this PR: I think we will merge it, and continue using mrpt-1.5.x for **released** ROS1 nodes.

bergercookie:
@jolting @jlblancoc I think ROS2 still has limited support for catkin-based CMake packages: 

https://index.ros.org/doc/ros2/Contributing/Migration-Guide/#continue-to-use-catkin-in-cmake

Although the transition catkin -> ament_cmake is relatively straightforward

jolting:
@bergercookie ROS2 is still a work in progress, so this might change, but that catkin dependency certainly gave me errors. MRPT2 built just fine on my setup without it.

As a 3rd party library we're not really using ament_cmake or catkin.

PR #912, opened 2019-04-13
add std designator before an endl

lsolanka:
## Changed apps/libraries

* graphslam

## PR Description

Small fix to add `std::` in front of an `endl` in a header file.

(Notify: @MRPT/owners )

Does changelog need to be updated?


jlblancoc:
Thanks! 
Actually, using `\n` would probably make more sense in general everywhere instead of endl, but let's leave that as low priority...

PR #921, opened 2019-05-24
Fix MacOS; refactor lightweight geometry types header

jlblancoc:
## Changed apps/libraries

* mrpt-math

## PR Description

* Refactor `<mrpt/math/lightweight_geom_data.h>` into a number of individual much smaller headers.
* Add base trivial types to TPoint2D, TPoint3D to allow their use inside dynamic matrices.


jlblancoc:
Merging, despite I couldn't fix the build on MacOS... The PR also contains an important refactoring of TPoseXX/TPointXX classes, finally putting some order in their definition headers.

PR #924, opened 2019-05-31
Remove custom aligned STL containers

### LINKED ISSUE #871: Use c++17 aligned new
See 3
https://en.cppreference.com/w/cpp/memory/new/operator_new

All classes must properly specify alignment requirements.

This probably makes the aligned*.h code unnecessary.

jlblancoc:
Fixed-size overaligned structures do not longer require special
allocators since C++17.

See, for example: http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1409

Closes #871


jolting:
Nice. That makes things simpler.

jlblancoc:
Yep!
However, it seems to have broken Xenial ppa builds in all architectures...
https://launchpadlibrarian.net/426268286/buildlog_ubuntu-xenial-amd64.mrpt_1%3A1.9.9~snapshot20190531-1857-git-76914136-xenial-1~ppa1~xenial_BUILDING.txt.gz

Could you please take a look and let me know what do you think?
I could only think of a possible bug in that gcc version?

jolting:
Or eigen that ships with xenial. I'll see what I can find.

jolting:
Yep. Eigen version. `EIGEN_USE_EMBEDDED_VERSION:BOOL=ON` works.

jlblancoc:
Great, thanks for testing!

However, I'm quite frustrated now: apparently, PPA builders no longer allow downloads, so eigen3 cannot be added at build time as a external project as it used it be. 

I guess the option is to add another PPA dependency to enable an eigen3 backported, only to Xenial... (sigh) I couldn't find any suitable such PPA in a first look, will come back to this later.

jolting:
Have you used the `backportpackage` tool?

jlblancoc:
Wow! I was thinking of backporting it by hand, you saved me a lot of time 👏👏

jlblancoc:
In the end I had to backport it by hand (a direct backport complained about deps...), but it was easy. Now we have eigen-3.3.7 for Xenial in the MRPT PPA, and Xenial builds are happy again :+1:

Issue #669, opened 2018-03-02
Make pf-localization work with either SE(2) or SE(3) poses


jlblancoc:
Right now, it's hard wired to work with mrpt::slam::CMonteCarloLocalization2D, but it could be refactored (templatized) so it could work with mrpt::slam::CMonteCarloLocalization3D as well. 
The choice should be done via a config file parameter.

jolting:
Nice. How well does it perform on a drone?

jlblancoc:
Hmm... interesting :-) 
Haven't tested it. Only tested for now with a ground vehicle + VLP-16. I have an "old" dataset of a dron with a velodyne onboard, could give it a try....

Issue #714, opened 2018-04-17
Turn some apps into C++ classes so they can be reused in unit tests & mrpt_performance


jlblancoc:
This will allow cleaning up duplicated code, adding new tests, and increasing code coverage.

List of candidates: 
* [x] pf-localization
* [x] icp-slam
* [x] icp-slam-live
* [x] rbpf-slam
* [x] graph-slam-engine (already done?)
* [x] map-partition -> not worth. Using regular unit tests of the underlying class instead.
* [x] grid-matching
* [x] kf-slam
* [x] rawlog-grabber

jolting:
Also for reuse in future Web UIs.

PR #925, opened 2019-06-03
pf-localization: add support for 3D poses (Closes: #669)

jlblancoc:


jlblancoc:
Merging, despite the spaghetti code of pf-localization hurts... Have no time to refactor it now. 
Connected to #714

PR #931, opened 2019-06-23
Fix obvious CFeatureLines warning

jolting:
Does this crash?

segments_out[i].size() is a constexpr, so no overhead.

---
This change is [Reviewable](https://reviewable.io/reviews/mrpt/mrpt/931)


jlblancoc:
Oh... that's why code coverage matters. Thanks!

PR #932, opened 2019-06-23
 Include <ostream>, <iosfwd> doesn't instantiate function on macOS

jolting:
This change is [Reviewable](https://reviewable.io/reviews/mrpt/mrpt/932)


jlblancoc:
Wow, great hunting!
 I am on the road today and couldn't check it out, but if CI is happy, we all are so ;-) 👏

PR #934, opened 2019-06-30
Small vector optimization

jlblancoc:
Add SVO to matrix classes directly, instead of using the vendor-dependant std::basic_string.

---
This change is [Reviewable](https://reviewable.io/reviews/mrpt/mrpt/934)

### DIFF
--- a/libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
+++ b/libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
@@ -161,15 +161,14 @@ class vector_with_small_size_optimization
 
 	const_reference back() const
 	{
-		return m_is_small ? m_a.back() : m_v.back();
+		return m_is_small ? m_a[m_size - 1] : m_v.back();
 	}
-	reference back() { return m_is_small ? m_a.back() : m_v.back(); }
+	reference back() { return m_is_small ? m_a[m_size - 1] : m_v.back(); }
 
 	const_reference front() const
 	{
 		return m_is_small ? m_a.front() : m_v.front();
 	}
-
 	reference front() { return m_is_small ? m_a.front() : m_v.front(); }
 
 	void swap(vector& x)
--- a/libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
+++ b/libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
@@ -30,7 +30,7 @@ struct UnspecializedBool
  * \note In `#include <mrpt/containers/vector_with_small_size_optimization.h>`
  * \ingroup mrpt_containers_grp
  */
-template <typename VAL, size_t small_size, size_t alignment>
+template <typename VAL, size_t small_size, size_t alignment = 16>
 class vector_with_small_size_optimization
 {
    private:
@@ -69,6 +69,7 @@ class vector_with_small_size_optimization
 		using pointer = POINTER;
 		using iterator_category = std::random_access_iterator_tag;
 		using difference_type = typename large_vec::difference_type;
+		iteratorImpl() = default;
 		iteratorImpl(STORAGE ptr) : m_ptr(ptr) {}
 		self operator++()
 		{
@@ -121,7 +122,7 @@ class vector_with_small_size_optimization
 		bool operator!=(const self& o) { return m_ptr != o.m_ptr; }
 
 	   private:
-		STORAGE m_ptr;
+		STORAGE m_ptr{nullptr};
 	};
 
 	using iterator = iteratorImpl<VAL, VAL*, VAL&>;
@@ -135,18 +136,21 @@ class vector_with_small_size_optimization
 			{
 				m_v.assign(m_a.begin(), m_a.begin() + m_size);
 			}
-			else if (!m_is_small && n < small_size)
+			else if (!m_is_small && n <= small_size)
 			{
 				std::copy(m_v.begin(), m_v.begin() + n, m_a.begin());
 			}
 		}
 		m_size = n;
-		m_is_small = n < small_size;
+		m_is_small = (n <= small_size);
 		if (!m_is_small)
 		{
 			m_v.resize(m_size);
 		}
 	}
+
+	size_t size() const { return m_size; }
+	bool empty() const { return m_size == 0; }
 
 	reference operator[](size_type n) { return m_is_small ? m_a[n] : m_v[n]; }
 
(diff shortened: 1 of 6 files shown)


jolting:
@jlblancoc I found out my laptop was thermal throttling during tests. Not the best thing to benchmark with. However, I think one of the map resize benchmarks might have been improved by this.

libs/math/include/mrpt/math/CMatrixDynamic.h
```diff
@@ -22,6 +22,162 @@
...
+			}
+			else if(!m_is_small && n < small_size)
+			{
+				std::copy (m_v.begin(), m_v.begin()+n, m_a.begin() );
+			}
+		}
+		m_size = n;
+		m_is_small = n < small_size;
+		if(!m_is_small){
+			m_v.resize(m_size);
```
jlblancoc: Shouldn't this resize() go before the assign() on line 124 above?

jlblancoc:
Well, test added in fb4b77ab6fd9599bda265853280c42ba70744618

The gain is still marginal (perhaps the compiler is optimizing out something since the matrices are not actually being used?), but it's something, I would vote to keep the svo thing: 

current master: 
```
mrpt-performance -c "matrix: vector"

matrix: vector of, resize(10) dyn[double] 4x4               1.99 us
matrix: vector of, resize(100) dyn[double] 4x4              7.98 us
matrix: vector of, resize(1000) dyn[double] 4x4             81.06 us
matrix: vector of, resize(10) dyn[double] 5x5               809.80 ns
matrix: vector of, resize(100) dyn[double] 5x5              8.46 us
matrix: vector of, resize(1000) dyn[double] 5x5             80.69 us
matrix: vector of, resize(10) fix[double] 4x4               140.50 ns
matrix: vector of, resize(100) fix[double] 4x4              301.30 ns
matrix: vector of, resize(1000) fix[double] 4x4             2.56 us
matrix: vector of, resize(10) fix[double] 5x5               201.30 ns
matrix: vector of, resize(100) fix[double] 5x5              906.70 ns
matrix: vector of, resize(1000) fix[double] 5x5             8.44 us
```

this PR:
```
matrix: vector of, resize(10) dyn[double] 4x4               2.04 us
matrix: vector of, resize(100) dyn[double] 4x4              8.22 us
matrix: vector of, resize(1000) dyn[double] 4x4             75.40 us
matrix: vector of, resize(10) dyn[double] 5x5               863.00 ns
matrix: vector of, resize(100) dyn[double] 5x5              7.36 us
matrix: vector of, resize(1000) dyn[double] 5x5             79.65 us
matrix: vector of, resize(10) fix[double] 4x4               146.90 ns
matrix: vector of, resize(100) fix[double] 4x4              298.10 ns
matrix: vector of, resize(1000) fix[double] 4x4             2.38 us
matrix: vector of, resize(10) fix[double] 5x5               207.00 ns
matrix: vector of, resize(100) fix[double] 5x5              898.90 ns
matrix: vector of, resize(1000) fix[double] 5x5             8.72 us
```

(Timing for "fixed" matrices shouldn't change between the tests, changes are just "random")

libs/math/include/mrpt/math/CMatrixDynamic.h
jolting: You mean a reserve() to avoid the potential reallocation?

jlblancoc: Hmm... nope, sorry, I didn't mean line 124 actually, it's line 128. 
If I got it right, it might be that we get into line 128 and do std::copy() **before** the m_v.resize(), with m_v having an arbitrary previous size (even empty)?

jolting: Copying from a large vector to the array doesn't require a resize. Both line 128 and line134 will never be a valid execution path. Arrays don't need a resize. Note line 132 sets m_is_small.

jlblancoc: (facepalm) Nervermind... got the arguments of std::copy() the other way around!

jlblancoc:
This is ready to be merged once CI is done.

libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
```diff
@@ -135,26 +136,42 @@ class vector_with_small_size_optimization
 			{
 				m_v.assign(m_a.begin(), m_a.begin() + m_size);
 			}
-			else if (!m_is_small && n < small_size)
+			else if (!m_is_small && n <= small_size)
```
jolting: Good catch!!

libs/containers/include/mrpt/containers/vector_with_small_size_optimization.h
```diff
@@ -161,15 +161,14 @@ class vector_with_small_size_optimization
 
 	const_reference back() const
 	{
-		return m_is_small ? m_a.back() : m_v.back();
+		return m_is_small ? m_a[m_size - 1] : m_v.back();
```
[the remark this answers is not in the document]
jlblancoc: I had to change this to fix a regression in tests for CVectorDouble and push_back()... if you think about it a bit it makes sense after all. It's undefined for empty containers, just like the std.

So good the body of unit tests grows.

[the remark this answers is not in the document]
jolting: I was just thinking about the trivial case where `m_size = 0` then `front() == back()`. `front()` should be m_a[0] always. I think you want `m_a[m_size]`, not `m_a[m_size - 1]` because `m_a[m_size]` is `m_a[0]` when the array is empty. Did I miss something?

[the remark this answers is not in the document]
jlblancoc: > front() should be m_a[0] always.
Sure!

> I think you want m_a[m_size], not m_a[m_size - 1]
Hmm... that would be where `end()` points to, but `back()` is the ref to the last valid element... isn't it?
I agree in that the trivial case with `m_size==0` would lead to `size_t(-1)`, and I wasn't comfortable with that, until I saw that for STL containers, it's undefined behavior to call `back()` when empty, so I saw it was acceptable to just... leave it "undefined behavior" (actually, would be a crash). 

We could add an ASSERT_() for debug builds, on a second thought...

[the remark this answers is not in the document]
jolting: Right. I was thinking of end(). Thx. I got it now.

jlblancoc:
Merged, thanks @jolting for the great alternative to `std::basic_string`, with well-defined, predictable behavior. Cool!

PR #1002, opened 2019-12-04
Request CMake to find OpenCV 2

jerry73204:
## Changed apps/libraries

- `cmakemodules/script_opencv.cmake`

## PR Description

This patch encourages CMake to find OpenCV 2 by default. Since I have multiple OpenCV installed. Especially Arch Linux allows installing `opencv` (>=4) and `opencv2-opt` in the mean time. The change makes sure build script will not find the wrong version.

### DIFF
--- a/cmakemodules/script_opencv.cmake
+++ b/cmakemodules/script_opencv.cmake
@@ -25,7 +25,7 @@ endif()
 
 # 1st option: Try to find OpenCV config file (NO_MODULE: Don't find a module, but OpenCVConfig.cmake):
 if(NOT CMAKE_MRPT_HAS_OPENCV)
-	find_package(OpenCV QUIET NO_MODULE)
+	find_package(OpenCV 2 QUIET NO_MODULE)
 	if(OpenCV_FOUND)
 		set(MRPT_OPENCV_VERSION ${OpenCV_VERSION})
 		set(OpenCV_LIBRARIES ${OpenCV_LIBS})


jlblancoc:
Thanks! I'll do some more further checking locally before merging.

jlblancoc:
With this PR, mrpt fails to find opencv when the unique installed versions are opencv 3 or 4, which is not acceptable... 
@jerry73204 what was the problem you wanted to solve with this change? 
If it's just selecting a specific opencv version, the standard way to do it is to set the OpenCV_DIR cmake variable, which is way more generic and less invasive.

jerry73204:
To elaborate, it's my assumption that MRPT links against OpenCV 2, even though 3 and 4 are present.
That's because I observed the MRPT includes `opencv2/*` headers.

If my assumption is wrong. It turns to another problem. In some distros, for example Arch Linux, the headers of OpenCV 4 is placed under `/usr/include/opencv4`. It would fail the compilation due to missing headers.

jlblancoc:
Ahh, ok thanks!
Have you checked whether there's an opencv2 directory under the headers directoryr opencv4?
I guess it should be like that, since opencv headers were named opencv2 after v1, but the name remained after v3, v4,...
In that case, cmake would handle the opencv4 prefix in the include path... that's my guess, I never tested mrpt in ArchLinux...

Could @cmorgan also check this?

jerry73204:
The [link](https://www.archlinux.org/packages/extra/x86_64/opencv/) contains the file listing of `opencv` on Arch (open link and click "View the file list for opencv" below). It shows OpenCV 4 headers are placed under `/usr/include/opencv4/opencv2/`.

jlblancoc:
Brilliant! Then everything is as expected, and one can choose the opencv version with cmake without further changes. I was thinking on packaging mrpt for Arch once v2.0.0 is finally released, so these checks are more than welcome. 
Cheers

PR #1125, opened 2021-01-26
Update the ExprTk library

ArashPartow:

### DIFF
--- a/libs/expr/include/mrpt/3rdparty/exprtk.hpp
+++ b/libs/expr/include/mrpt/3rdparty/exprtk.hpp
@@ -2,7 +2,7 @@
  ******************************************************************
  *           C++ Mathematical Expression Toolkit Library          *
  *                                                                *
- * Author: Arash Partow (1999-2018)                               *
+ * Author: Arash Partow (1999-2021)                               *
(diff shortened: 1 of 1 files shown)


jlblancoc:
All relevant unit tests passed, merging. 

Thanks! :+1:

jlblancoc:
Just in case this was a bug (?), please check these lines:

```
/home/jlblanco/code/mrpt/libs/expr/include/mrpt/3rdparty/exprtk.hpp:1931:29: warning: declaration of ‘exponent’ shadows a previous local [-Wshadow]
                   const int exponent = static_cast<int>(-std::distance(curr, itr));
                             ^~~~~~~~
/home/jlblanco/code/mrpt/libs/expr/include/mrpt/3rdparty/exprtk.hpp:1904:14: note: shadowed declaration is here
          int exponent = 0;
              ^~~~~~~~
```

ArashPartow:
It's a different exponent. It's essentially the power to 10 the fractional part needs to be multiplied with to make it an integer.

I'll rename it so as to make it less confusing and so as not to trigger that diagnostic. 

Thank-you for raising this issue.

jlblancoc:
ok, thanks for clarifying :+1:

PR #1146, opened 2021-03-04
Make MRPT_EXCEPTIONS_WITH_CALL_STACK public

jolting:
MRPT_EXCEPTIONS_WITH_CALL_STACK was not public,
so it wasn't being defined in unit tests or apps.

Any define used in a public header file should be
a PUBLIC define.

---

I acknowledge to have:

(Notify: @MRPT/owners )


jolting:
Just realized this is wrong.

jlblancoc:
hehe, no problem! It's only used in one .cpp file, so it should remain private... 

Also, I don't want to "polute" the list of arguments of downstream projects, so as a general rule we should prefer `config.h` files to cmake `PUBLIC` definitions.

PR #1160, opened 2021-08-05
CImage: Fix joinImagesHorz through exception

anhtuk54:
1) With empty image an exception will be through, eg:
CDisplayWindow::showImagesAndMatchedPoints
  -> CImage::joinImagesHorz
   -> CImage::getChannelCount
    -> CImage::makeSureImageIsLoaded
       Run into "else if (!allowNonInitialized)" case
       and through exception.
2) After resize, ano ther cv::Mat is created and assign to
m_impl->img. Therefore "img" will not point to correct current
cv::Mat anymore.

This commit fixes these issue and makes CDisplayWindow::showImagesAndMatchedPoints
works normally as expected.


jlblancoc:
LGTM, thanks! :+1:

PR #1214, opened 2022-02-02
Update ExprTk

ArashPartow:


jlblancoc:
Thanks, @ArashPartow ! 

This goes through our unit tests, but just in case: any new feature or backwards incompatible change worth knowing and/or documenting?

ArashPartow:
@jlblancoc: There are no interface changes that will affect MRPT.  All the changes are internal implementation.

PR #1328, opened 2024-12-06
Rename TBoundingBox None Ctor flag

J-Meyers:
## Changed apps/libraries

* Renamed flag to bounding box from None to NoFlags to prevent autogenerated stubs from conflicting with python None type

## PR Description

- Fixes python typing issue where previously it tried to annotate another type to the None type in the autogenerated stub
  - This issues is particularly visible if trying to use mypy which generates an unrecoverable error (line number difference is because this was done using an older commit of mrpt)
  - The current issue effectively prevents using mypy for type checking in any projects which have any sort of transitive import of mrpt
image

---

I acknowledge to have:
* Updated [`doc/source/doxygen-docs/changelog.md`](https://github.com/MRPT/mrpt/blob/develop/doc/source/doxygen-docs/changelog.md) to describe these changes (if applicable)

(Notify: @MRPT/owners )


jlblancoc:
Good point! It would be great to scan the rest of the Python wrapping code for other similar cases...

PR #1374, opened 2026-07-02
feat(math): KDTreeCapable save/load of the KD-tree index (2D and 3D)

jlblancoc:
## Summary
- Port from mrpt-2.x (MRPT/mrpt#1373): adds `kdtree_save_index_2D/3D()` and `kdtree_load_index_2D/3D()` to `mrpt::math::KDTreeCapable`, allowing a built KD-tree index to be serialized/restored without rebuilding (e.g. to cache alongside a serialized point cloud). Only the tree topology is stored; a point-count header is validated on load.
- Adds `MRPT_HAS_KDTREE_SAVE_LOAD_INDEX` feature-detection macro.
- Adds unit tests in `modules/mrpt_maps/tests/CPointsMap_unittest.cpp` using `CSimplePointsMap` (not present in the original 2.x commit): 2D/3D round-trip, empty-map save behavior, and point-count-mismatch error handling.

## Test plan
- [x] `colcon build --packages-up-to mrpt_math` builds cleanly
- [x] `colcon build --packages-up-to mrpt_maps` builds cleanly
- [x] `test_mrpt_maps --gtest_filter="CSimplePointsMapTests.kdtree*"` — 4/4 new tests pass
- [x] Full `CSimplePointsMapTests.*` / `CGenericPointsMapTests.*` suite — 19/19 pass

## Summary by CodeRabbit

* **New Features**
  * Added the ability to save and load KD-tree search indexes for both 2D and 3D point maps.
* **Bug Fixes**
  * Improved reliability when reusing a map after queries in another dimension.
  * Saving now safely handles empty maps.
  * Loading validates the stored point count and errors on mismatches.
* **Tests**
  * Added unit tests covering successful reload with matching point counts, empty-map behavior, and point-count mismatch handling.

### DIFF
--- a/modules/mrpt_maps/src/maps/CReflectivityGridMap2D.cpp
+++ b/modules/mrpt_maps/src/maps/CReflectivityGridMap2D.cpp
@@ -161,7 +161,7 @@ bool CReflectivityGridMap2D::internal_insertObservation(
     *cell = static_cast<cell_t>(cell_new);
 
     return true;  // Done!
-  }  // end if "CObservationGasSensors"
+  }               // end if "CObservationGasSensors"
 
   return false;
 
--- a/modules/mrpt_maps/tests/CPointsMap_unittest.cpp
+++ b/modules/mrpt_maps/tests/CPointsMap_unittest.cpp
@@ -21,6 +21,7 @@
 
 #include <array>
 #include <sstream>
+#include <stdexcept>
 
 using namespace mrpt;
 using namespace mrpt::maps;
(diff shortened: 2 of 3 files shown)


jlblancoc:
Addressed all 3 CodeRabbit findings in the amended commit:
1. `rebuild_kdTree_2D/3D()` now also check that the per-dimension index pointer is non-null (not just the shared `m_kdtree_is_uptodate` flag), so querying one dimension then saving the other no longer spuriously returns `false`. Added a regression test (`kdtreeSaveIndexAfterOtherDimensionQueried`).
2. Save methods now assert the output stream is still good after each write.
3. Load methods now build the new index into a local `unique_ptr` and only commit it to the member state (clearing stale trees, setting `m_kdtree_is_uptodate`) after the header validation and `loadIndex()` have both succeeded, so a failed load can no longer leave the object in a corrupted 'up-to-date but null index' state.

PR #1375, opened 2026-07-02
fix(system): guard COutputLogger's callback list against concurrent register/log

jlblancoc:
## Summary
- `COutputLogger::m_listCallbacks` was read (`logStr()`, iterating over it to invoke callbacks) and written (`logRegisterCallback()` / `logDeregisterCallback()`) without any synchronization, even though these can legitimately happen from different threads -- e.g. a GUI thread hooking a live logger via `logRegisterCallback()` while that logger's own thread is concurrently calling `logStr()`. That's a data race / UB, not just a theoretical edge case.
- Guards `m_listCallbacks` with a `shared_ptr<mutex>`, mirroring the existing `m_historyMtx` pattern (a plain `std::mutex` member would make `COutputLogger` non-copy-assignable, breaking `loggerReset()`'s `*this = COutputLogger()`).
- `logStr()` copies the callback list under the lock and invokes callbacks *outside* it, so a slow/blocking user callback can't stall a concurrent registration, and a callback that itself calls `logRegisterCallback()`/`logDeregisterCallback()` on the same logger can't deadlock.

## Context
Found while hardening [MOLAorg/mola](https://github.com/MOLAorg/mola)'s new `mola_viz_imgui` Console log window, which periodically discovers already-running modules and registers a log callback on each from a dedicated GUI thread -- exactly the race this fixes.

## Test plan
- [x] `libs/system` builds clean with the change.
- [x] Downstream consumer (`mola_viz_imgui`, which derives from `COutputLogger`) rebuilds and runs correctly against the patched library (verified via a throwaway repro harness registering callbacks on live, actively-logging loggers from a separate thread).
- [ ] Existing MRPT `system` unit tests / CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Dnw8SX9MWvxpwoTgkR155K

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread safety for logging callbacks, reducing the risk of crashes or missed notifications when callbacks are added, removed, or triggered at the same time.
  * Logging now processes callback notifications more reliably under concurrent use.

### DIFF
--- a/libs/system/include/mrpt/system/COutputLogger.h
+++ b/libs/system/include/mrpt/system/COutputLogger.h
@@ -338,6 +338,13 @@ class COutputLogger
   mutable std::deque<TMsg> m_history;
   std::shared_ptr<std::mutex> m_historyMtx = std::make_shared<std::mutex>();
 
+  /** Guards m_listCallbacks: logRegisterCallback()/logDeregisterCallback() may
+   * be called (e.g. by a GUI thread hooking into a logger) concurrently with
+   * logStr() running on the logger's own thread, which iterates the list.
+   * A shared_ptr, like m_historyMtx above, so COutputLogger stays copy-
+   * assignable (std::mutex itself is not; loggerReset() relies on this via
+   * `*this = COutputLogger()`). */
+  std::shared_ptr<std::mutex> m_listCallbacksMtx = std::make_shared<std::mutex>();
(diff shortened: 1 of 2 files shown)


jlblancoc:
@coderabbitai review

PR #1386, opened 2026-08-20
test(mrpt_core): regression test for Clock::toDouble unsigned wraparound

jlblancoc:
## Summary

`mrpt::Clock::toDouble()` on the 2.x maintenance line has a bug: it computes
the epoch subtraction with a `uint64_t` operand, which promotes the signed
`int64_t` tick count to unsigned and wraps around for any timestamp before
1970-01-01. See the companion fix on 2.x: #TBD (opening right after this PR).

**On `develop`, `toDouble()` is already correct.** It does the subtraction in
the floating-point domain:

```cpp
return (static_cast<double>(t.time_since_epoch().count()) - 116444736e9) / 10000000.0;
```

This double-precision subtraction never wraps, so no code change is needed
here. I verified this myself before touching anything (see the standalone
reproduction below) rather than "fixing" code that isn't broken.

What *is* missing on `develop` is a regression test that would catch this
class of bug. The existing `fromDouble_negativeAndNearEpoch` test (added by
b9e4174a, which fixed the sibling UB in `fromDouble()`) only checks a
*difference* of two `toDouble()` values:

```cpp
EXPECT_NEAR(mrpt::Clock::toDouble(tm1) - mrpt::Clock::toDouble(tm2), 1e-3, 1e-9);
```

A difference cancels a `2^64` wraparound, so this test passes whether or not
`toDouble()` itself is correct — it would not have caught the 2.x bug had it
been present here too.

## Changes

- `modules/mrpt_core/src/Clock.cpp`: no behavior change. Gave `toDouble()`
  the same named `UNIX_EPOCH_OFFSET` constant that `fromDouble()` already
  uses (added by b9e4174a), to make the epoch arithmetic self-documenting on
  both sides and keep the two functions easy to compare. The subtraction is
  still done as `double - double`, bit-identical to before (verified: a
  floating-point literal and `static_cast<double>` of the same integer value
  round to the same `double` under IEEE-754 round-to-nearest).
- `modules/mrpt_core/tests/Clock_unittest.cpp`: added
  - `toDouble_preEpoch`: a **single** pre-epoch `toDouble()` value, checked
    directly (not via a difference) — this is the test that would have
    caught the 2.x bug.
  - `toDouble_monotonicAcrossEpoch`: `toDouble()` must be strictly
    increasing across the 1601→1970 epoch boundary.
  - `fromDouble_toDouble_roundtrip_negativeZeroPositive`: round trip for
    negative, zero, and positive UNIX timestamps.

## Verification

Standalone `g++` reproduction of the exact `toDouble()` body used on
`develop` (floating-point subtraction) vs. the buggy unsigned-integer body
still present on 2.x, using the tick values from the bug report:

```
=== epoch - 2ms (ticks=116444735999980000) ===
  buggy : toDouble = 1844674407370.953125
  fixed : toDouble = -0.002000
=== epoch + 0.625s (ticks=116444736006250036) ===
  buggy : toDouble = 0.625004
  fixed : toDouble = 0.625003
=== difference (a - b), a=epoch-2ms, b=epoch+0.625s ===
  buggy : diff = 1844674407370.328125  (should be -0.627, wraps to ~ -1.8e12)
  fixed : diff = -0.627003  (expected -0.627)
=== post-epoch (t >= 1970) bit-identical check: PASS (identical for all t >= 1970 samples) ===
```

Confirms `develop`'s existing implementation is already correct, and that
the difference-based test alone would not distinguish buggy from fixed.

## Downstream report

`MOLAorg/mola_state_estimation#60` — `toDouble()`'s pre-epoch wraparound
(present on 2.x, not on develop) caused a spurious `-1.8e12 s` time
difference and an assertion failure that reset the state estimator on 22 of
22 KITTI runs, because `mola_lidar_odometry` backdates its first two fused
poses by 1-2 ms and KITTI's first scan is stamped `0.000000`.

Companion fix on the 2.x line (the branch that actually ships): see the PR
targeting `2.x` opened alongside this one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected timestamp conversion for dates before 1970.
  * Preserved accurate ordering and round-trip behavior across the UNIX epoch, including small negative timestamps.

* **Tests**
  * Added regression coverage for negative, zero, and positive timestamp conversions.

### DIFF
--- a/apps/mrpt_apps_cli/carmen2rawlog/carmen2rawlog_main.cpp
+++ b/apps/mrpt_apps_cli/carmen2rawlog/carmen2rawlog_main.cpp
@@ -36,6 +36,7 @@
 
 #include <CLI/CLI.hpp>
 #include <fstream>
+#include <iostream>
 #include <map>
 
 using namespace mrpt;
--- a/apps/mrpt_apps_cli/carmen2simplemap/carmen2simplemap_main.cpp
+++ b/apps/mrpt_apps_cli/carmen2simplemap/carmen2simplemap_main.cpp
@@ -39,6 +39,7 @@
 
 #include <CLI/CLI.hpp>
 #include <fstream>
+#include <iostream>
 
 using namespace mrpt;
 using namespace mrpt::poses;
--- a/apps/mrpt_apps_cli/gps2rawlog/gps2rawlog_main.cpp
+++ b/apps/mrpt_apps_cli/gps2rawlog/gps2rawlog_main.cpp
@@ -27,6 +27,7 @@
 #include <mrpt/system/os.h>
 
 #include <CLI/CLI.hpp>
+#include <iostream>
 
 using namespace mrpt;
 using namespace mrpt::hwdrivers;
--- a/modules/mrpt_core/src/Clock.cpp
+++ b/modules/mrpt_core/src/Clock.cpp
@@ -203,7 +203,17 @@ double mrpt::Clock::toDouble(const mrpt::Clock::time_point t) noexcept
     return .0;  // invalid time point
   }
 
-  return (static_cast<double>(t.time_since_epoch().count()) - 116444736e9) / 10000000.0;
+  // Offset between the 1601-01-01 epoch used by Clock::duration and the
+  // 1970-01-01 UNIX epoch, in 100-nanosecond ticks.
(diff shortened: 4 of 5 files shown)


jlblancoc:
Companion fix on the 2.x line (the branch that actually ships to users): #1387

PR #1388, opened 2026-08-28
test(mrpt_nav, mrpt_kinematics): raise coverage to 90%/97% and fix 10 bugs found on the way

jlblancoc:
## Summary

Raises unit-test coverage of the navigation stack, targeting the two largest
pure-logic blind spots left in the per-module table of `agents.md`:

| Module | Lines before | Lines after | Branches before | Branches after |
|---|---|---|---|---|
| `mrpt_nav` | 63.0% | **90.0%** | 45.3% | 69.3% |
| `mrpt_kinematics` | 38.2% | **96.8%** | 17.9% | 83.5% |

`mrpt_kinematics` had **no C++ unit tests at all** (only a Python bindings
smoke test); `mrpt_nav`'s were limited to a handful of reactive-navigation
integration runs plus a few data-structure checks.

Measured with the recipe documented in `agents.md` (whole-repo
`colcon build -DENABLE_COVERAGE=ON` + `colcon test` + `gcovr`, deduplicated
with `scripts/coverage_module_report.py`).

## New test files

**`mrpt_kinematics`** (new `UNIT_TEST_SOURCES` list — the module had none):
- `CVehicleVelCmd_unittest.cpp` — both kinematic models: component
  accessors and their out-of-range throws, stop detection, scaling,
  `cmdVel_limits()` clipping/blending, serialization, `TVelCmdParams`
  config round-trip.
- `CVehicleSimul_unittest.cpp` — differential-driven and holonomic
  simulators: straight/rotational motion, local vs. global velocities, the
  first-order delay model, odometry error injection, reset semantics, and
  the wrong-kinematic-class rejection paths.
- `CKinematicChain_unittest.cpp` — link management, DH forward kinematics
  (revolute and prismatic), origin pose, configuration vectors,
  serialization, and the 3D visualization helpers.

**`mrpt_nav`**:
- `PTG_variants_unittest.cpp` — the whole PTG family, including
  `CPTG_DiffDrive_CC` / `_CS` / `_CCS`, which no test had ever instantiated
  (~2% line coverage each), plus the shared base services: config and stream
  round-trips, TP-space index mapping, clearance diagrams, `debugDumpInFiles()`,
  path/robot-shape rendering, and the legacy numeric PTG names.
- `CAbstractNavigator_unittest.cpp` — the navigation state machine driven by
  a recording mock robot: every state transition, the error paths, the
  blocked-target and not-approaching alarms, and the waypoint-sequence layer
  (skip policy, non-skippable waypoints, look-ahead, heading alignment).
- `TWaypoint_unittest.cpp` — waypoint text/config/visualization round-trips.
- `nav_misc_unittest.cpp` — `ClearanceDiagram`, the arc-vs-circular-robot
  collision helper, the default `CRobot2NavInterface` callbacks,
  `NavigationLogger`, `VelocityFilter` and the multi-objective optimizers.
- `planners_and_logs_unittest.cpp` — the RRT move-tree 3D renderer
  (`impl_renderMoveTree.h`, previously 0%), planner configuration variants,
  and a fully-populated `CLogFileRecord` round-trip.
- `holonomic_config_unittest.cpp` — config/serialization/log-record coverage
  for `CHolonomicVFF` / `ND` / `FullEval`.
- `rnav_variants_unittest.cpp` — reactive-navigation runs from fully
  in-memory configs (so they cannot silently skip when the shared config
  files are absent), exercising the delays model, clearance evaluation,
  velocity filtering, disabled obstacle filtering, log-record keeping and
  file writing, and — via a `CPTG_Holo_Blend` — the "NOP cmdvel"
  PTG-continuation branches that the differential-drive PTGs cannot reach.

Both modules now meet the 90% line-coverage goal stated in `agents.md`.
294 new/extended test cases across 11 files; all 78 module test suites pass.

## Bugs found and fixed

1. **`CVehicleVelCmd` copy constructor called a pure virtual method.**
   It delegated to `operator=()`, which dispatches `getVelCmdLength()` /
   `setVelCmdElement()` while the derived object is still under
   construction. Copy-constructing *any* velocity command aborted the
   process (`pure virtual method called`). The base class holds no data, so
   the copy constructor is now `= default`.

2. **`CAbstractNavigator` threw on the first step of every waypoint mission
   and every relative-target navigation.**
   `internal_onStartNewNavigation()` cleared the cached pose history but left
   `m_last_curPoseVelUpdate_robot_time` untouched, so the
   `updateCurrentPoseAndSpeeds()` call right after it was skipped by its
   20 ms minimum-period throttle and the cache stayed empty; the following
   `ASSERT_(!m_latestPoses.empty())` then threw. The exception was swallowed
   (see the next item), so the symptom was a silently lost first navigation
   step.

3. **`NAV_ERROR` could never be reached.**
   `performNavigationStepNavigating()` ended with
   `m_navigationState = prevState;`, restoring the state it had on entry and
   undoing the transitions it had just decided — neither an exception nor
   `doEmergencyStop()` could leave the navigator in `NAV_ERROR`. The
   assignment was clearly meant for `m_lastNavigationState`, mirroring
   `navigationStep()` a few hundred lines above.

4. **Null-pointer dereference in `CWaypointsNavigator::checkHasReachedTarget()`.**
   `(wp == nullptr && targetDist <= allowed) || (wp->reached)` dereferences
   `wp` whenever it is null *and* the robot is farther than the allowed
   distance — reachable e.g. after emptying the list through
   `getWaypointsAccessGuard()`.

5. **`CParameterizedTrajectoryGenerator::Alpha2index()` discarded
   `wrapToPi()`'s result** (it returns by value), so directions outside
   `[-pi,pi]` were clamped to the first or last path instead of wrapping
   around.

6. **`CMultiObjectiveMotionOptimizerBase::decide()` returned `-1` from a
   function returning `std::optional<size_t>`** when a user score/assert
   formula failed to compile. That is an *engaged* optional holding
   `SIZE_MAX`, so instead of the documented "no valid candidate" the caller
   indexed the candidate vector out of bounds.

7. **`CMultiObjectiveMotionOptimizerBase::clear()` left the variable table
   behind**, so the next `decide()` threw *"Expression name already exists
   as an input variable"* — i.e. the documented way to re-apply changed
   parameters was unusable.

8. **`CLogFileRecord`'s legacy (pre-v15) deserialization wrote velocity
   command components into the wrong slots**: one loop indexed with the
   outer loop variable, and the oldest format wrote both `v` and `w` into
   element 0.

9. **`CReactiveNavigationSystem3D::saveConfigFile()` never called its parent.**
   Unlike its 2D sibling, and contrary to the documented contract
   ("derived classes must call the parent implementation"), so saving a 3D
   navigator's configuration produced a stub with only `HEIGHT_LEVELS` and
   `PTG_COUNT` that could not be loaded back.

10. **`PlannerSimple2D::computePath()`'s endpoint bounds check was inverted.**
    `!(originInside || !targetInside)` flags exactly one of the four cases,
    and the wrong one: it only reported `notFound` when the origin was outside
    *and* the target inside, letting an out-of-grid target (or both endpoints
    out of grid) fall through into the grid search.

## Observations left for a follow-up (not changed here)

- `CWaypointsNavigator::m_last_alignment_cmd` is set in the constructor and
  never updated when an alignment command is issued, so the
  "give the alignment some time to finish" wait actually measures time since
  the navigator was constructed. Fixing it would add a real dwell at every
  waypoint with a heading, which is a behavioral change better made
  deliberately.
- With a `CPTG_Holo_Blend`, a waypoint left at the default
  `speed_ratio = 1.0` produces no viable movement at all and the mission
  eventually times out; `rnav_variants_unittest.cpp` sets an explicit
  `0.05` and notes this in a comment.
- `CKinematicChain::recomputeAllPoses()` advertises a `pose0` argument that
  the implementation marks `[[maybe_unused]]` and ignores in favour of
  `setOriginPose()`. Documented as such here rather than changing behavior;
  removing the parameter would be an API break.
- `agents.md`'s coverage table, overall figure and notes are updated to match,
  including the gotchas hit while writing these tests (the explicit
  `LIB_UNIT_TEST_SOURCES` list, `rnav_unittest.cpp` silently skipping when the
  shared config files are absent, the navigation-time throttle in
  `updateCurrentPoseAndSpeeds()`, and the per-PTG robot-shape requirements).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation state transitions, pose updates, waypoint handling, angle normalization, legacy log loading, and configuration clearing.
  - Corrected navigation configuration persistence and velocity-command copying.

- **Tests**
  - Added extensive coverage for kinematics, navigation, planners, trajectory generators, waypoints, reactive navigation, interfaces, logging, serialization, and configuration behavior.

- **Documentation**
  - Clarified how kinematic-chain origin poses are applied.

- **Chores**
  - Refreshed coverage metrics and module tracking information.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -201,7 +201,9 @@ files needed changes.
 A full rebuild of all 33 `modules/*` packages was done with coverage
 instrumentation, followed by a full `colcon test` run (all tests passed) and a
 `gcovr` line/branch report. **Goal: 90% line coverage per module.** Current
-overall (2026-07-06): **51.3% lines / 38.1% branches** — well short of goal.
+overall (2026-08-28, deduplicated the same way as
+`scripts/coverage_module_report.py`): **68.5% lines / 48.8% branches**
+— still short of goal, dominated by the hardware/GUI modules below.
 Per-module rows below are still the 2026-07-03 baseline except where a row is
 tagged with a newer date (e.g. `mrpt_math`, `mrpt_graphs`, after their
 unit-test passes).
@@ -292,14 +294,12 @@ and accurate path — pick two.
 | mrpt_gui | 22/4621 | 0.5% | 0.2% |
 | mrpt_hwdrivers | 913/6592 | 13.9% | 9.7% |
 | mrpt_comms | 237/1013 | 23.4% | 11.2% |
-| mrpt_kinematics | 184/482 | 38.2% | 17.9% |
 | mrpt_graphslam | 257/611 | 42.1% | 37.3% |
 | mrpt_opengl | 2035/4234 | 48.1% | 30.1% |
 | mrpt_system | 1012/1900 | 53.3% | 39.6% |
 | mrpt_io | 726/1292 | 56.2% | 39.9% |
 | mrpt_libapps_cli | 1130/1910 | 59.2% | 38.0% |
 | mrpt_libapps_gui | 952/1567 | 60.8% | 42.6% |
-| mrpt_nav | 3928/6234 | 63.0% | 45.3% |
 | mrpt_slam | 2778/4299 | 64.6% | 43.7% |
 | mrpt_viz (2026-08-02) | 6449/9426 | 68.4% | 50.1% |
 | mrpt_rtti | 126/176 | 71.6% | 73.5% |
@@ -308,6 +308,7 @@ and accurate path — pick two.
 | mrpt_math (2026-07-05) | 6914/8070 | 85.7% | 57.5% |
 | mrpt_core | 541/628 | 86.1% | 64.8% |
 | mrpt_obs (2026-07-10) | 4631/5324 | 87.0% | 56.1% |
+| mrpt_nav (2026-08-28)¶ | 4524/5026 | 90.0% | 69.3% |
 | mrpt_img (2026-07-10)† | 2255/2495 | 90.4% | 69.2% |
 | mrpt_graphs (2026-07-06) | 1022/1111 | 92.0% | 76.7% |
 | mrpt_poses | 6263/6787 | 92.3% | 59.8% |
@@ -315,6 +316,7 @@ and accurate path — pick two.
 | mrpt_expr | 93/100 | 93.0% | 60.2% |
 | mrpt_random | 160/167 | 95.8% | 85.1% |
 | mrpt_bayes (2026-07-09) | 1036/1078 | 96.1% | 77.4% |
+| mrpt_kinematics (2026-08-28)¶ | 426/440 | 96.8% | 83.5% |
 | mrpt_config (2026-07-09) | 531/548 | 96.9% | 82.3% |
 | mrpt_tfest (2026-07-07) | 633/652 | 97.1% | 73.3% |
 | mrpt_topography (2026-07-10) | 373/375 | 99.5% | 91.2% |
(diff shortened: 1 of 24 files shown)


jlblancoc:
Both review findings addressed in 8f0ee8e:

1. **`CMultiObjectiveMotionOptimizerBase` partial expression state** — confirmed and fixed. Both failure paths now go through `clear()`. Verified the new regression test `MultiObjOpt.a_failed_compilation_leaves_no_stale_state_behind` actually reproduces the bug: with the old `m_score_exprs.clear()` it fails with *"Expression name `aaa_ok` already exists as an input variable"*, and passes with the fix.

2. **Hard-coded `/tmp`** — replaced with `mrpt::system::extractFileDirectory(mrpt::system::getTempFileName())`. (The pre-existing `waypoints_and_rrt_unittest.cpp` still hard-codes it; left alone to keep this PR's diff scoped, worth a separate cleanup.)

PR #1389, opened 2026-08-28
test(graphslam, system, io, slam): cover the largest untested pure-logic files, fix 8 bugs

jlblancoc:
## Summary

Second coverage pass, following #1388. Targets the four largest remaining
**pure-logic** gaps (i.e. excluding `mrpt_hwdrivers`/`mrpt_gui`/`mrpt_imgui`,
which need a mockable transport layer or a live display):

| Module | Lines | Branches |
|---|---|---|
| `mrpt_graphslam` | 41.0% → **72.5%** | 33.0% → **59.6%** |
| `mrpt_slam` | 64.1% → **67.5%** | 43.1% → **44.9%** |
| `mrpt_io` | 57.0% → **63.6%** | 40.2% → **46.4%** |
| `mrpt_system` | 54.4% → **61.7%** | 40.5% → **41.9%** |

+453 newly covered lines. Measured with the recipe in `agents.md`
(whole-repo `-DENABLE_COVERAGE=ON` + full `colcon test` + `gcovr`,
deduplicated via `scripts/coverage_module_report.py`). All 78 module test
suites pass.

**Five files that had never had a single assertion run against them** are now
covered: `TSlidingWindow.cpp`, `CEdgeCounter.cpp`, `md5.cpp` and
`CRejectionSamplingRangeOnlyLocalization.cpp` (all at 0%), plus
`vector_loadsave.cpp` (16%).

## Bugs found and fixed

Each was reproduced by a failing test *before* the fix.

1. **`TSlidingWindow::getMean()` returned NaN on an empty window.**
   It divided by `m_measurements_vec.size()` without the emptiness guard that
   `getMedian()` has. The NaN then propagates: since every comparison against
   a NaN is false, `evaluateMeasurementAbove()` is stuck at `false` for *any*
   input, however large.

2. **`TSlidingWindow::getStdDev()` normalized by the window capacity**, not by
   the number of measurements held. A window of size 100 holding 8 samples
   whose true population σ is exactly `2.0` reported ≈`0.57`. It also
   disagreed with `getMean()`, which divides by the sample count.

3. **`TSlidingWindow::resizeWindow()` never invalidated the cached σ.** It
   resets the mean and median flags only — so σ is stale after a shrink, and
   on a *grow* the old code invalidated nothing at all, even though the
   reported value depended on `m_win_size`.

4. **`CEdgeCounter::clearAllEdges()` left `m_unique_edges` behind.** Every
   other counter is reset, so a "cleared" instance still reported a stale
   unique-edge total in `getAsString()`.

5. **`mrpt::system::md5(const std::vector<uint8_t>&)` was undefined behavior
   on an empty vector.** `&str[0]` indexes out of bounds, and the resulting
   pointer then tripped the `ASSERT_(data)` of the raw-buffer overload it
   delegates to — so `md5(empty_vector)` *threw* while `md5(empty_string)`
   correctly returned `d41d8cd98f00b204e9800998ecf8427e`. The algorithm
   itself passes every RFC 1321 published test vector; only the wrapper was
   wrong.

6–8. **`mrpt::io::vectorNumericFromTextFile()` — three defects in one
   function:**
   ```cpp
   while (!feof(f)) {
     int readed = fscanf(f, byRows ? "%lf" : "%lf\n", &number);
     if ((!byRows) || (readed == 1))   // readed ignored when !byRows
       vec.push_back(number);
   }                                    // FILE handle never closed
   ```
   - the `fscanf` return value is **discarded** in the default path, because
     `(!byRows)` short-circuits — a failed read still pushes the stale
     `number`, so an **empty file yielded `{0.0}`**;
   - the output vector is **never cleared**, so it appends to whatever the
     caller passed, unlike `loadTextFile()`/`loadBinaryFile()` which both
     `clear()` first;
   - the `FILE*` is **leaked** — every sibling function in the same file
     calls `os::fclose(f)`.

## A non-result worth recording

`CRejectionSamplingRangeOnlyLocalization.cpp` was the largest 0%-coverage file
in the repository (117 lines). The 10 new tests cover the beacon-height
projection onto the robot plane, sensor-offset compensation, the two-circle
intersection geometry, the angle-prefilter toggle and the error paths — and
**all passed first try**. It was untested, not broken.

## Left documented rather than changed

Both are genuinely ambiguous, so they are pinned down by tests describing
current behavior and flagged here for a maintainer decision:

- `CEdgeCounter::addEdge(name, is_loop_closure=true, is_new=true)` **throws**
  if the edge type already exists, but silently **drops the loop-closure
  count** if the type is new. Same arguments, opposite behavior depending on
  prior state. Either the throw is over-strict or the new-type branch should
  count it; the intent isn't recoverable from the code.
- `CControlledRateTimer`'s header documents its low-pass filter as
  `estimation = a0*input + (1-a0)*former_estimation`, but the implementation
  applies those weights the other way round. The code is the sensible reading
  for a *low*-pass filter (a0=0.9 ⇒ heavy smoothing), so the doc looks like
  the mistake — but the PI controller was presumably tuned against the
  current behavior, so changing either is a control-dynamics decision.

## Also fixed: a flaw in the documented coverage recipe

`agents.md`'s reproduce command uses `--base-paths modules`, which excludes
`apps/`. `rawlog-edit` is therefore never built and the ~40
`RawlogEditCLITest` cases all `GTEST_SKIP()` with *"rawlog-edit binary not
found"* — so following the recipe verbatim measures `mrpt_libapps_cli` at
**~8%** rather than the ~59% recorded in the table. Noted next to the recipe.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected sliding-window statistics for empty data, resizing, and standard deviation calculations.
  * Fixed edge-counter reset behavior.
  * Improved numeric vector file loading by replacing existing contents and ignoring failed reads.
  * Prevented unsafe handling of empty inputs when calculating MD5 hashes.

* **Documentation**
  * Corrected timer parameter defaults and low-pass filter behavior descriptions.

* **Tests**
  * Added comprehensive automated coverage for graph-slam, file I/O, SLAM localization, MD5, and timer functionality.
  * Updated coverage guidance and results documentation.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -674,11 +674,17 @@ Left documented rather than changed, as the intent is genuinely ambiguous:
 `CEdgeCounter::addEdge(name, is_loop_closure=true, is_new=true)` throws when
 the edge type already exists but silently drops the loop-closure count when
 it is new -- same arguments, opposite behavior depending on prior state.
-`CControlledRateTimer`'s header documents its low-pass filter as
-`estimation = a0*input + (1-a0)*former_estimation`, but the implementation
-applies those weights the other way round; the code is the sensible reading
-for a low-pass filter, so the doc looks like the mistake, but the PI
-controller was presumably tuned against the current behavior.
+
+`CControlledRateTimer`'s header docs were corrected to match the
+implementation rather than the reverse: the low-pass filter is
+`estimation = a0*former_estimation + (1-a0)*input` (the doc had the weights
+swapped), and the documented defaults for `Ti` (0.0194) and `a0` (0.9)
+disagreed with the member initializers (0.1 and 0.99). The control law itself
+was left untouched -- it is the sensible reading for a *low*-pass filter and
+the PI gains were presumably tuned against it. A new
+`CControlledRateTimer_unittest.cpp` pins both the documented defaults and the
+filter's weighting (with `a0 == 1` the estimate must ignore the measurement
+entirely and stay on the set-point), so the two cannot drift apart again.
 
 ### Weak areas, grouped by root cause
 
@@ -704,7 +710,7 @@ controller was presumably tuned against the current behavior.
 4. **Quick wins — pure-logic files at 0% with no hardware/GUI dependency**
    (highest-value gaps, ordinary unit tests would work immediately):
    `mrpt_graphslam/src/CWindowObserver.cpp`,
-   `mrpt_system/src/{CFileSystemWatcher,CControlledRateTimer}.cpp`.
+   `mrpt_system/src/CFileSystemWatcher.cpp`.
    (`mrpt_system/src/md5.cpp`, `mrpt_graphslam/src/{CEdgeCounter,
    TSlidingWindow}.cpp` and
    `mrpt_slam/src/slam/CRejectionSamplingRangeOnlyLocalization.cpp` all
--- a/agents.md
+++ b/agents.md
@@ -225,6 +225,13 @@ gcovr --root . -j$(nproc) --gcov-ignore-parse-errors=all \
   --exclude '.*_unittest\.cpp' --exclude '.*/python_bindings/.*' --exclude '.*/samples/.*' \
   --json-pretty -o coverage.json build
 ```
+Gotcha (2026-08-28): `--base-paths modules` deliberately excludes `apps/`, so
+the `rawlog-edit` binary is never built and the ~40 `RawlogEditCLITest` cases
+in `mrpt_libapps_cli` all `GTEST_SKIP()` with "rawlog-edit binary not found".
+Following the recipe verbatim therefore measures that module at ~8% rather
+than the ~59% recorded below. Add `apps` to the `--base-paths` list when the
+`mrpt_libapps_*` numbers matter.
+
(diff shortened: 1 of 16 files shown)


jlblancoc:
Follow-up in 56219c5: the `CControlledRateTimer` doc/impl mismatch flagged above is now **fixed in the docs**, per maintainer decision — the control law is untouched.

Three corrections to the header:
- the low-pass filter is `estimation = a0*former_estimation + (1-a0)*input`; the doc had the weights swapped, which inverted the meaning of `a0` (larger values smooth *more*, not less)
- `Ti` default is **0.1**, documented as 0.0194
- `a0` default is **0.99**, documented as 0.9

A new `CControlledRateTimer_unittest.cpp` (8 tests) pins both halves together so they cannot drift again: it asserts the documented defaults, and asserts that with `a0 == 1` the estimate ignores the measurement entirely and stays on the set-point — under the previously-documented formula it would instead equal the raw measured rate, so the test discriminates between the two readings. Parameter validation and `setRate()` are covered as well.

That also clears `CControlledRateTimer.cpp` (34 lines, was 0%) from the quick-wins list in `agents.md`.

PR #1404, opened 2026-09-05
fix(img): CStereoRectifyMap scrambled the forward axis for horizontal-baseline stereo pairs

jlblancoc:
## Bug

`CStereoRectifyMap::setFromCamParams()` derives the new rectified-frame axes `(e1, e2, e3)` from the baseline direction `e1` by crossing it with a fixed reference "up" vector to get `e2`, then sets `e3 = e1 x e2`. For an *ordinary side-by-side stereo rig* — where the baseline is itself close to the camera's own x axis — this assigns the original camera's z axis ("forward") to `e2` instead of `e3`. Since `build_rectify_map()` unprojects each rectified pixel by dividing by the z-component of the rotated ray, every pixel of every image from a normal stereo pair rectified this way divides by ~0 and maps outside the source image, producing a blank (or near-blank) rectified output.

Verified independently in Python (replicating the algorithm) and in C++: a 3D point directly in front of the original left camera at z=3 transforms to `(0, 3, 0)` under the old `e1/e2/e3` — its entire depth lands on the y axis instead of staying on z.

No existing test caught this: `Rectify_basic`/`Rectify_color` only assert output image *dimensions*, using a smooth gradient input that survives any remap, garbled or not.

## Fix

Derive `e3` (not `e2`) from the reference vector instead, via a Gram-Schmidt projection of the camera's own forward axis `(0,0,1)` orthogonalized against the baseline, falling back to `(0,1,0)` only when the baseline is itself nearly colinear with the optical axis (a forward-facing, depth-separated pair). `e2 = e3 x e1` keeps `(e1, e2, e3)` right-handed, matching the camera's own x-right/y-down/z-forward convention (x cross y = z).

## Test

New `CStereoRectifyMap.Rectify_preservesForwardAxis`: with a plain 0.10 m x-baseline pair and zero relative rotation, a marker at the left camera's principal point must reappear near the rectified image's own principal point. Confirmed this test fails against the old code (the whole rectified output is black, `sumW == 0`) and passes with the fix. All 12 pre-existing `CStereoRectifyMap`/`CUndistortMap` tests still pass.

Found while wiring a real fisheye stereo rig (GrandTour's Alphasense front_left/front_right) through `mola_vision`'s stereo VO, which expects a pre-rectified pair.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DErBjLRU438kY9bKWYAaWo

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stereo image rectification to preserve the camera’s forward/depth direction.
  - Corrected geometric alignment near the image center, improving rectified feature placement and reducing projection errors.
  - Improved rectified image positioning for challenging baseline and optical-axis orientations.
  - Rectified camera geometry now consistently represents a positive horizontal baseline without unintended rotation.

- **Tests**
  - Added coverage for relative camera rotation, ideal rectification, accurate feature placement, and near-degenerate baseline orientations.


jlblancoc:
Thanks - both findings addressed in `1bcd81a2a`:

1. **Forward reference not rotated into the half-rotated frame.** Fixed exactly as suggested: `e3` is now derived from `R_half * z` (falling back to `R_half * y`) instead of the bare, unrotated axis, projected orthogonal to `e1` before normalizing.
2. **Degeneracy threshold too coarse.** Switched from the `|e1.dot(z)| > 0.9` angle check to the actual projected-norm check, so a baseline like `(sqrt(1-0.91^2), 0, 0.91)` no longer gets routed through the fallback unnecessarily.

Added the two requested regression cases:
- `Rectify_preservesForwardAxis_withRelativeRotation` - a real non-identity `R_half` (5 deg relative yaw), checking the same principal-point invariant as the original test.
- `Rectify_obliqueBaselineNearOpticalAxis` - the oblique-baseline case from the review. This one checks that `setFromCamParams()`/`rectify()` succeed and that the rectification rotation is a valid, finite unit quaternion - not that a principal-point marker survives, since a baseline this close to the optical axis is a "forward motion" stereo configuration where even correct rectification legitimately reprojects the principal point outside the original field of view (verified numerically: the source ray for this exact case lands at pixel x=-777 on a 640px-wide camera - a real geometric property of this configuration, not a defect).

All 15 `CStereoRectifyMap` tests and the 9 `CUndistortMap` tests pass.

PR #1405, opened 2026-09-06
Coverage pass on mrpt_graphslam and mrpt_gui (headless GUI tests under Xvfb)

jlblancoc:
Raises `mrpt_graphslam` 73.3% → 81.6% and `mrpt_gui` **0.5% → 40.6%** lines (whole repo: 73.8% → 76.4%), and fixes the bugs the new tests uncovered.

## Headless GUI tests

`mrpt_gui` had no `tests/` directory at all. It has one now: `xvfb-run` plus Mesa's software rasterizer is enough to open and drive `CDisplayWindow`, `CDisplayWindow3D` and `CDisplayWindowPlots` with no real display, which carried `mathplot.cpp` 0%→33%, `WxSubsystem.cpp` 4.7%→67%, `CDisplayWindow3D.cpp` 0%→74%, `CDisplayWindow.cpp` 0%→71%, `CDisplayWindowPlots.cpp` 0%→65%.

* `.github/workflows/build-linux.yml` installs `xvfb`/`libgl1-mesa-dri` and wraps `colcon test` in `xvfb-run`.
* `modules/mrpt_gui/package.xml` gains `<test_depend>xvfb</test_depend>` (documentation/rosdep only — the CI wrapper is what actually makes the tests run).
* Every window test goes through `SKIP_IF_NO_GUI()`, so the suite still passes with no display; `MRPT_SKIP_GUI_TESTS=1` forces that path.

Deliberately **not** doing pixel comparisons of window screenshots: on a virtual display there is no compositor, the window is never mapped, and reading its pixels back gives a uniformly black image even though rendering happened. The tests assert that a frame was grabbed with a plausible size; reference-image comparisons stay in `mrpt_opengl`'s offscreen EGL/FBO tests, which need no display.

Validated locally both under `xvfb-run` (with `DISPLAY` unset) and on a real X session; full `colcon test` over `modules` + `apps` is green in both.

**Please watch the macOS and Windows jobs**: those runners have a native window server, so the window tests will actually run there rather than skip. If they turn out to be unhappy, the gate in `tests/gui_test_common.h` is the single place to restrict them to Linux.

## Bugs found and fixed

* **`CTicTac` over-alignment → segfault in every optimized build.** `alignas(16)` on storage that is only ever reinterpreted as `struct timespec`/`LARGE_INTEGER` propagated through `CTimeLogger` up to `CRegistrationDeciderOrOptimizer`, which is used as a *virtual base*. GCC then compiled its members assuming a 16-byte-aligned `this` and emitted `movdqa`, while the virtual-base subobject inside a derived class sits at an 8-byte-aligned offset — so merely constructing a `CFixedIntervalsNRD` crashed in Release, i.e. `graphslam-engine` was broken. Same failure mode already documented in `mrpt::math::CMatrixFixed`; a `static_assert` now pins it.
* **Root node placed at the wrong pose.** `CNodeRegistrationDecider::registerNewNodeAtEnd()` seeded the root with `getCurrentRobotPosEstimation()` instead of the origin, applying the pre-registration motion twice (node 1 landed at 2× the travelled distance).
* **Uninitialized registration thresholds** in all three decider `TParams` (`CFixedIntervalsNRD`, `CIncrementalNodeRegistrationDecider`, `CICPCriteriaNRD`): the documented defaults existed only inside `loadFromConfigFile()`, so a decider used without `loadParams()` compared against garbage.
* **Two templates that could not be instantiated at all**: `CIncrementalNodeRegistrationDecider` (undefined `report_sep`, unqualified `INVALID_NODEID`) and `TUncertaintyPath::hasLowerUncertaintyThan()` (a `const` method calling the non-`const` `getDeterminant()`; the cache is now `mutable`).
* **`CDisplayWindow3D::getLastMousePosition()` always returned pixel (0,0)** — and so did `getLastMousePositionRay()` and all 3D picking. `CGlCanvasBase::updateLastPos()` was declared but never defined after the 3.x camera refactor, and `CWxGLCanvasBase`'s mouse handlers never recorded the pointer position. Also defines `CGlCanvasBaseHeadless::renderError()`, whose absence made that documented class unlinkable; the two dead click setters (superseded by `COrbitCameraController`) are removed.
* **`wxImage2MRPTImage()` swapped red and blue** — a leftover from MRPT 2.x when `CImage` stored BGR. In 3.x both sides are RGB.
* **`CWindowObserver` missed Ctrl+C** whenever another modifier was held (`key_modifiers == 8192` instead of masking `MRPTKMOD_CONTROL`).
* **`using namespace std;` at global scope in a public header** (`CNodeRegistrationDecider_impl.h`), leaking into every TU that included any NRD.

`agents.md` is updated with the new numbers, the Xvfb recipe and the gotchas.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved 3D mouse-position tracking for more reliable picking.
  - Corrected image conversions to preserve RGB channel order.
  - Added safe default registration thresholds for graph-slam components.
  - Ensured newly created graph roots start at the origin.
  - Improved Ctrl+C handling and diagnostic logging.
  - Prevented potential alignment issues in timing utilities.

- **Tests**
  - Expanded GUI and graph-slam test coverage, including headless GUI execution in CI.
  - Added coverage for window behavior, image conversion, registration, uncertainty paths, and event handling.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -60,6 +60,10 @@ jobs:
                            libcli11-dev \
                            libzstd-dev
 
+          # Virtual X display + Mesa software GL, so that the mrpt_gui window
+          # tests can run headlessly (they self-skip without a display):
+          sudo apt install -y xvfb libgl1-mesa-dri
+
           # Added: python3-pip and colcon installation via pip
           sudo apt install -y pybind11-dev python3-dev python3-pip
           pip3 install --break-system-packages colcon-common-extensions colcon-lcov-result numpy
@@ -76,9 +80,12 @@ jobs:
 
       - name: Colcon Test
         run: |
-          colcon test \
-            --event-handlers console_direct+ \
-            --return-code-on-test-failure
+          # `xvfb-run` provides the virtual display the mrpt_gui tests need;
+          # every other test is unaffected by it.
+          xvfb-run -a --server-args="-screen 0 1280x1024x24" \
+            colcon test \
+              --event-handlers console_direct+ \
+              --return-code-on-test-failure
 
       - name: Generate Coverage Report
         if: matrix.coverage == 'ON'
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -3,6 +3,11 @@ name: CI Linux
 
 on: [push, pull_request]
 
+# Nothing here writes to the repository: checkout only reads, and Codecov
+# uploads with its own token.
+permissions:
+  contents: read
+
 jobs:
   build:
     name: ${{ matrix.name }} ${{ matrix.build_type }}
--- a/agents.md
+++ b/agents.md
@@ -201,12 +201,17 @@ files needed changes.
 A full rebuild of all 33 `modules/*` packages was done with coverage
 instrumentation, followed by a full `colcon test` run (all tests passed) and a
 `gcovr` line/branch report. **Goal: 90% line coverage per module.** Current
-overall (2026-08-31, deduplicated the same way as
-`scripts/coverage_module_report.py`): **73.6% lines / 54.4% branches**
-- still short of goal, dominated by the hardware/GUI modules below.
-The whole table below was re-measured on 2026-08-31; the date tags on some
+overall (2026-09-06, deduplicated the same way as
+`scripts/coverage_module_report.py`): **76.4% lines**
+- still short of goal, dominated by the hardware modules below.
+The whole table below was re-measured on 2026-09-06; the date tags on some
 rows mark when that module last had a dedicated unit-test pass, and point at
 the footnote describing it.
+
+**The GUI tests need a display**: run `colcon test` under
+`xvfb-run -a --server-args="-screen 0 1280x1024x24"` (as CI does), or the
+`mrpt_gui` window tests `GTEST_SKIP()` and that module reads ~1% instead of
+~40%. `MRPT_SKIP_GUI_TESTS=1` forces the skip.
 
 Gotcha (2026-07-06): the system `gcov` alias (`/etc/alternatives/gcov`) may
 point to an unrelated binary (observed pointing to `/usr/bin/gc`), causing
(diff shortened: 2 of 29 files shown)


jlblancoc:
Thanks — went through all three inline comments. Two applied, one declined.

**1. `build-linux.yml` — restrict `GITHUB_TOKEN` (applied).** Added top-level `permissions: contents: read`. Verified nothing in the workflow needs write: `actions/checkout` only reads, and the Codecov step uploads with `secrets.CODECOV_TOKEN`, not the workflow token.

**2. `CTicTac.h` — `largeInts` under-aligned for `LARGE_INTEGER` (applied; valid, and a regression this PR introduced).** `unsigned long` is 4 bytes and 4-byte aligned on MSVC, so removing the `alignas(16)` left the storage 4-byte aligned while `LARGE_INTEGER` (containing a `LONGLONG`) needs 8. Linux masked it — `unsigned long` is 8 bytes there — which is why CI stayed green.

Fixed by changing the storage to `std::uint64_t largeInts[4]`, whose natural alignment is 8 on every platform. That satisfies both `LARGE_INTEGER` and `struct timespec` without any `alignas`, and 8 is still well below the over-alignment that caused the original virtual-base crash. The constructor now asserts alignment as well as size, separately per platform.

Re-verified locally that this does not undo the original fix: `alignof` of `CTicTac`/`CTimeLogger`/`CRegistrationDeciderOrOptimizer` is 8, and the `CFixedIntervalsNRD` construction repro that used to segfault exits cleanly at both `-O2` and `-O3`. Full `colcon test` over `modules` + `apps` is green (30 suites, 0 failures).

**3. `gui_test_common.h` — probe the display server before running GUI tests (declined).** The premise is accurate: `CBaseGUIWindow` waits 6 s (30 s in debug) and then leaves the window unopened, so a stale `DISPLAY` gives slow failures rather than skips. Declining because:

- It cannot happen in CI: the Linux jobs run under `xvfb-run` (a working display is guaranteed), and the macOS/Windows jobs build with `-DDISABLE_WXWIDGETS=ON` and skip earlier.
- Actively probing would mean an `XOpenDisplay`/socket connect for X11 plus a separate mechanism for Wayland — platform-specific code and a new dependency in a test helper, for a misconfigured local dev environment only.
- Failing is arguably the right signal there. Turning "the display is broken" into a silent skip would also hide a genuine regression that stopped windows opening.

Instead I documented the situation in the helper, along with the two existing opt-outs: `MRPT_SKIP_GUI_TESTS=1` to skip, and `MRPT_WXSUBSYS_TIMEOUT_MS` to shorten the wait.

PR #1410, opened 2026-09-08
mrpt_nav: modernize the PTG/reactive API and fix 7 TP-Space bugs

jlblancoc:
An API and mathematics pass over `mrpt_nav`, separate from the recent coverage work (#1388). Test count 246 → 251; all pass.

## API modernization

`bool` + out-param signatures replaced by `std::optional`, following what `inverseMap_WS2TP()` already does. The old forms stay as `[[deprecated]]` shims unless noted:

- `CParameterizedTrajectoryGenerator::getPathStepForDist(k, dist)`. The old 3-arg form *also* wrote the last path step into `out_step` when it returned `false`, and two call sites quietly depended on that; the explicit `getPathStepForDistClamped()` now covers it.
- `nav_plan_geometry_utils::collision_free_dist_{segment,arc}_circ_robot()`
- `PlannerSimple2D::computePath()`

Other changes:

- `ClearanceDiagram::getClearance()`'s `bool integrate_over_path` → `enum class ClearanceQuery`. One flag had three different names across the codebase (`integrate_over_path`, "spot, dont interpolate", "interpolate") and none matched the implementation — see bug 1.
- `CPTG_Holo_Blend::PATH_TIME_STEP`, a process-wide mutable global backing a *per-instance* virtual (`getPathStepDuration()`), becomes the per-instance `path_time_step` config key + `setPathTimeStep()` (serialization v5). `eps` becomes `constexpr EPSILON`. Neither admits a compatible shim.
- `setScorePriorty()` → `setScorePriority()`.
- Deleted: `updateClearancePost()`, a documented no-op since 2017, and `CAbstractHolonomicReactiveMethod::Create()` — declared in a public header but **never defined anywhere**, so any call was a link error.

## Bugs fixed

Each has a regression test that fails without the fix.

1. **`getClearance()`'s two modes were swapped** relative to its own docs, to every call site's comment, and to the ptg-configurator's UI label: `false` averaged over the path and `true` returned the spot value. The reactive navigator's `clearance` and `clearance_path` score factors therefore held each other's values (only reachable with `evaluate_clearance=true`, off by default).
2. **`initClearanceDiagram()` keyed samples by raw path distance in metres** while every consumer treats those keys as normalized [0,1] TPS distances: `getClearance()` queries, the `dist_over_path > 0.5` collision heuristic, and `CAbstractPTGBasedReactive`'s own `dist_eucl_min` producer, which builds the same map keyed `i/num_steps`.
3. Same function sampled steps `0, incr, 2·incr…` while `evalClearanceSingleObstacle()` evaluates `incr, 2·incr…`, so every clearance value was filed under a shorter distance than the one it was measured at.
4. **`CHolonomicVFF::navigate()` assigned `desiredSpeed` inside `if (m_enableApproachTargetSlowDown)`**, so with the slow-down disabled it returned speed 0 and the robot never moved. `CHolonomicND`/`CHolonomicFullEval` had it right.
5. `CHolonomicFullEval` slowed down for `targets.front()`; `NavInput` documents the *last* target as the highest-priority one.
6. **`CPTG_Holo_Blend` used `V_MAX` as the post-ramp cruise speed** in `getPathDist()`, `getPathStepForDist()` and `updateTPObstacleSingle()`, while `getPathPose()` advances at the direction-dependent `internal_get_v(dir)`. With an `expr_V` configured, poses and distances disagreed. `inverseMap_WS2TP()` likewise pinned `T_ramp = T_ramp_max` instead of `internal_get_T_ramp(alpha)`; it is now re-evaluated per Newton iteration (the Jacobian omits dT/dα, which costs iterations, not accuracy, since the residual is exact).
7. **`collision_free_dist_arc_circ_robot()` returned NaN for any obstacle on the turn-center axis** — the closed form divides by the obstacle's `x`. Rewritten as a two-circle intersection; it agrees with the old formula to 1.7e-11 over 21k random collision cases, and now returns 0 when the robot starts already in collision instead of the *exit* distance.

## Math model

`CPTG_Holo_Blend`'s arc-length quadrature moves from a 15-interval trapezoidal rule to 16-interval Simpson — the same number of function evaluations for ~25× lower mean relative error — plus an exact branch for the degenerate `b²−4ac ≈ 0` case, where the integrand is `sqrt(a)·|t−r|`.

## Docs and examples

The PTG base class description is rewritten around what TP-Space actually is, dropping its 20-year change log; `inverseMap_WS2TP()`'s docs still listed out-params removed in 3.x. New headless example `mrpt_examples_cpp/nav_ptg_tpspace` walks the whole WS → TP-Space → velocity-command round trip.

`agents.md` §11 records the whole pass.

## Summary by CodeRabbit

- **New Features**
  - Added a headless console example for TP-Space path planning, obstacle handling, clearance queries, and velocity commands.
  - Added configurable per-instance path timing for blended holonomic trajectories.
  - Added optional-result APIs for path planning, collision checks, and trajectory sampling.
  - Added explicit clearance query modes and clamped trajectory-step lookup.

- **Bug Fixes**
  - Improved clearance calculations, collision detection, speed control, and trajectory distance integration.
  - Fixed edge cases involving empty or decimated paths and direction-dependent speeds.

- **Documentation**
  - Documented navigation API updates and added example usage guidance.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -1298,3 +1298,72 @@ command sender each `dynamic_cast`ed to `CSerialPort` and asserted on it, so
 any other stream aborted on the first read. Those four functions only ever call
 `Read()`/`Write()`, so they now use the bound stream; `open`/`setConfig`/
 `purgeBuffers` are genuinely serial-specific and were left alone.
+
+## 11. mrpt_nav API modernization + TP-Space math fixes (2026-09-08)
+
+An API/mathematics pass over `mrpt_nav`, separate from the coverage pass of ¶.
+Nothing here is a coverage-driven change; the test count went 246 -> 251.
+
+**Out-param APIs replaced by `std::optional`** (old signature kept as a
+`[[deprecated]]` inline shim unless noted), matching what `inverseMap_WS2TP()`
+already did:
+
+* `CParameterizedTrajectoryGenerator::getPathStepForDist(k, dist)`. The old
+  3-arg form also wrote the *last* path step into `out_step` when it returned
+  false, and two call sites silently relied on that; the explicit
+  `getPathStepForDistClamped()` now covers it.
+* `nav_plan_geometry_utils`: `collision_free_dist_{segment,arc}_circ_robot()`.
+* `PlannerSimple2D::computePath()`.
+
+**Other API changes**: `ClearanceDiagram::getClearance()`'s `bool
+integrate_over_path` became `enum class ClearanceQuery` (see below);
+`CPTG_Holo_Blend::PATH_TIME_STEP` (a mutable global) became the per-instance
+`path_time_step` config key + `setPathTimeStep()`, and `eps` became
+`EPSILON`, both no-shim breaks; `setScorePriorty()` -> `setScorePriority()`;
+`updateClearancePost()` (a documented no-op since 2017) and
+`CAbstractHolonomicReactiveMethod::Create()` (declared in the header but
+**never defined anywhere** -- calling it was a link error) were deleted.
+
+**Real bugs fixed** -- each has a regression test that fails without the fix:
+
+1. `ClearanceDiagram::getClearance()`'s two modes were **swapped** relative to
+   its own docs, to every call site's comment, and to the ptg-configurator's
+   UI label; `bool=false` averaged over the path and `bool=true` returned the
+   spot value. The reactive navigator's `clearance` / `clearance_path` score
+   factors were therefore each other's values (only visible with
+   `evaluate_clearance=true`, off by default).
+2. `initClearanceDiagram()` keyed the clearance samples by the raw path
+   distance **in meters** while every consumer treats those keys as normalized
+   [0,1] TPS distances (`getClearance()` queries, the `dist_over_path > 0.5`
+   collision heuristic, and `CAbstractPTGBasedReactive`'s own
+   `dist_eucl_min` producer, which builds the same map keyed `i/num_steps`).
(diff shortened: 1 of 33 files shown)


jlblancoc:
Thanks — all three findings were real; fixed in 888c837.

1. **`getPathStepForDist()` unit mismatch (major)** — confirmed and fixed. `d` is the min of two normalized quantities, while the API takes pseudometers; the ETA factor ~300 lines below does `d * ref_dist` for the very same call, and `CHolonomicFullEval` does the same. This is pre-existing (the PR only converted the call mechanically), and it meant the "end of trajectory" behind `robpose_*`, `dist_eucl_final` and the target slow-down check was read `ref_distance` times too early along the path. The new `RNavVariants.end_of_trajectory_pose_is_taken_at_the_right_path_distance` measures 0.74 m instead of 4 m without the fix. Also switched to `getPathStepForDistClamped()` as suggested: with the units fixed, a candidate running to the very end of the path could now trip the assertion.

2. **`m_pathStepCountCache` invalidation** — correct, the cache counts path *time steps*. Every write to `m_pathTimeStep` (`loadFromConfigFile()`, `loadDefaultParams()`, `serializeFrom()`, `setPathTimeStep()`) now goes through `setPathTimeStep()`, which clears it.

3. **Invalid serialized time steps** — folded into the same place: `setPathTimeStep()` rejects non-finite and non-positive values, so the deserialization path is covered by (2) rather than needing its own check.

252/252 tests pass.

